PR #51, opened 2015-01-22
Update PHYS14 dataset list.

geonmo:
These list can be used to run crab as parameter.

Recent PHYS14 List at 2015/1/20 is used.


jshlee:
+1

jhgoh:
Do we have expectation on the sample size?

jshlee:
for each sample, i think we need to add in the global tag to be used

jhgoh:
How about defining JSON or xml to bind dataset/reference cfg/cross section info?
(for a mid-term task, but let us not to go too much)

jshlee:
sounds good

jhgoh:
+1
This PR is just a configuration for the production and nothing to do on the code structure.
I think it is safe to merge without detailed review.

PR #59, opened 2015-01-28
minor changes to configs

jshlee:
also adding pesudo top as default


geonmo:
For test,

``` bash
git checkout -b testPR59
git pull origin pull/59/head:cat72x
```

CatProducer/python/catTemplate_cfg.py
```diff
@@ -31,23 +31,13 @@
...
-        #'keep *_TriggerResults_*_PAT',
-        #'keep *_patTrigger*_*_*',
-        #'keep *_*_*_PAT',
-    )
+    outputCommands = cms.untracked.vstring('drop *')
 )
+process.out.outputCommands += catEventContentExtended
+
+print process.out.outputCommands
+
```
monttj, line 38: This should be configured in the run configuration file. Please see the PAT2CAT_cfg.py.

CatProducer/python/catSetup_cff.py
```diff
@@ -39,8 +39,11 @@ def catSetup(process, runOnMC=True, doSecVertex=True):
         ## Load MC dependent producers
         process.load("CATTools.CatProducer.pdfWeight_cff")
         process.load("CATTools.CatProducer.pileupWeight_cff")
+        process.load("CATTools.CatProducer.pseudoTop_cfi")
         process.out.outputCommands.append("keep *_pdfWeight_*_*")
         process.out.outputCommands.append("keep *_pileupWeight_*_*")
+        process.out.outputCommands.append("keep *_pseudoTop_*_*")
+        process.out.outputCommands.append("keep *_partonTop_*_*")
```
monttj: Can we maintain the output in catEventContent__cff single file? so we don't need to look around here and there.

monttj:
Why don't we use automatic global tag for CAT? If it should be changed, it should be changed at up-level, PAT configuration file in CMSSW.

jshlee:
from what i've seen so far, the auto doesnt work so well
we can change the default to auto and override with globaltag input

CatProducer/python/catSetup_cff.py
jhgoh: This event content entry is only for MC. Wouldn't it crash in real data if we put all in static event content definition?

jshlee: im using it for data in cmssw5
the collections are there but empty

jshlee:
can i merge this pull request now???

DataFormats/interface/Muon.h
```diff
@@ -49,6 +49,32 @@ namespace cat {
...
+    float puChargedHadronIso(float dR=0.3) const {
+      if( dR < 0.35) return puChargedHadronIso03_;
+      else return puChargedHadronIso04_;
+    }
+    float neutralHadronIso(float dR=0.3) const {
+      if( dR < 0.35) return neutralHadronIso03_;
+      else return neutralHadronIso04_;
+    }
+    float photonIso(float dR=0.3) const {
+      if( dR < 0.35) return photonIso03_;
```
monttj, line 65: why dR < 0.35 instead of dR == 0.3 ?

DataFormats/interface/Electron.h
```diff
@@ -34,24 +34,20 @@ namespace cat {
...
-      else return -1.0;
+      if( dR < 0.35) return neutralHadronIso03_;
+      else return neutralHadronIso04_;
     }
     float photonIso(float dR=0.3) const {
-      if( dR == 0.3) { return photonIso03_;}
-      else if( dR == 0.4) { return photonIso04_;}
-      else return -1.0;
+      if( dR < 0.35) return photonIso03_;
+      else return photonIso04_;
```
monttj, line 50: same as muon

DataFormats/interface/Muon.h
jhgoh, line 65: To avoid possible truncating. Safe and no additional cost.

jshlee:
i'll merge this to being with and add in the other changes to the config files in a bit

PR #70, opened 2015-02-03
reverting back to individual candidate producers

jshlee:


jhgoh:
Just a quick question : wouldn't there any case to treat MC as a real data?
For example temporary turn off MC matching in some reason?

jshlee:
I dont see a case where that is used...
Can you think of any?

monttj:
+1

Issue #61, opened 2015-01-29
documentation


jshlee:
wiki maintenance
https://www.cms-kr.org/twiki/bin/view/Computing/CATTools/WebHome

jshlee:
i think our documentation is going well.
if anyone has any more suggestions for this, please make a new issue.

PR #78, opened 2015-02-11
Cat reco example

dhkim81:
This is fully working example of EDAnalyzer to run on catTuple.
Basic kinematic distributions of most of physics objects in the catTuple are produced.


dhkim81:
histogram_structure

dhkim81:
histogram_structure_2

jhgoh:
Reopened it.

jshlee:
if all variables are saved, wouldnt a tree be better than histos?

jhgoh:
This is aimed to be a fully working example, replacement of existing CATMuonAnalysis.cc
My idea is to give an example for beginners, but also to be used to validate productions as CMS release validation does.

We already have a generic tree producer, so we don't have to put those functions in this simple analyzer.

jshlee:
yep, thats fine then. 
can you get @dhkim81 and @hcwhwang to document these on the kcms wiki once the site is back up.

jshlee:
please add in the instructions in https://www.cms-kr.org/twiki/bin/view/Computing/CATToolsExampleWebHome

jhgoh:
Please move CATMuonAnalysis.cc to plugins directory.
src is not for the plugins. 

Still, I don't know why this dummy analysis module is needed...

jhgoh:
@jshlee Would it be OK to have instruction before merging into main branch?

jshlee:
yep, I would prefer documentation done at the same time

jhgoh:
(related to the issue #61 )

There are several places to put examples:
- (from you reply) https://www.cms-kr.org/twiki/bin/view/Computing/CATToolsExampleWebHome
- (the main page) https://www.cms-kr.org/twiki/bin/view/Computing/CATTools/WebHome#How%20To
- https://github.com/vallot/CATTools/blob/cat72x/CatAnalyzer/README.md

I think 3rd one can be removed.
Link in the 2nd bullet is mostly on dataformat, setup and production. 
Why don't we have analysis examples there?

jshlee:
I put https://www.cms-kr.org/twiki/bin/view/Computing/CATTools/WebHome#Analysis Examples 
to link to https://www.cms-kr.org/twiki/bin/view/Computing/CATToolsExampleWebHome
so that all analysis examples go in one place

jshlee:
@dhkim81 please add in instructions to this new location
https://www.cms-kr.org/twiki/bin/view/Computing/CATTools/AnalysisExamples/WebHome

PR #80, opened 2015-02-11
cleaned catEventContent and added catEventContentMC and catEventContentA...

jshlee:
cleaned catEventContent and added catEventContentMC and catEventContentAODMC


jhgoh:
+1

PR #82, opened 2015-02-12
fixed jet Uncertainty

jshlee:
still need to update the JES
easy to implement on the fly JEC later on

### DIFF
--- a/CatProducer/plugins/CATJetProducer.cc
+++ b/CatProducer/plugins/CATJetProducer.cc
@@ -48,6 +48,7 @@ namespace cat {
     const std::vector<std::string> btagNames_;
     std::string uncertaintyTag_, payloadName_;
     bool runOnMC_;
+    JetCorrectionUncertainty *jecUnc;
 
   };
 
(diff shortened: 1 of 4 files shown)


CatProducer/plugins/CATJetProducer.cc
```diff
@@ -78,36 +70,19 @@ cat::CATJetProducer::produce(edm::Event & iEvent, const edm::EventSetup & iSetup
...
-  edm::Handle<pat::JetCollection> smearedResDownSrc;
-  edm::Handle<pat::JetCollection> smearedResUpSrc;
-  if (runOnMC_){
-    iEvent.getByToken(shiftedEnDownSrc_, shiftedEnDownSrc);
-    iEvent.getByToken(shiftedEnUpSrc_, shiftedEnUpSrc);
-    iEvent.getByToken(smearedResSrc_, smearedResSrc);
-    iEvent.getByToken(smearedResDownSrc_, smearedResDownSrc);
-    iEvent.getByToken(smearedResUpSrc_, smearedResUpSrc);
+  edm::ESHandle<JetCorrectorParametersCollection> JetCorParColl;
+  if (payloadName_.size()){
```
jhgoh, line 74: almost ignorable comment:

``` cpp
if ( !payloadName_.empty()) {
```

is better.

CatProducer/python/jetProducer_cfi.py
```diff
@@ -2,12 +2,9 @@
...
     src = cms.InputTag("selectedPatJetsPFlow"),
-    shiftedEnDownSrc = cms.InputTag("shiftedSlimmedJetsEnDown"),
-    shiftedEnUpSrc = cms.InputTag("shiftedSlimmedJetsEnUp"),
-    smearedResSrc = cms.InputTag("smearedSlimmedJets"),
-    smearedResDownSrc = cms.InputTag("smearedSlimmedJetsResDown"),
-    smearedResUpSrc = cms.InputTag("smearedSlimmedJetsResUp"),
-    btagNames = cms.vstring()
+    btagNames = cms.vstring(),
+    ## payloadName should be AK4PFchs, but PHYS14_25_V2 does not have uncertainty 
+    payloadName = cms.string(""),
```
jhgoh, line 7: So, is this just a dummy pset? Isn't it crashing?

CatProducer/plugins/CATJetProducer.cc
jshlee, line 74: was in a rush. had to go drinking! hehe

CatProducer/python/jetProducer_cfi.py
jshlee, line 7: it doesnt crash
in PHYS14_25_V2 they didnt add the uncertainties, while it exists in PHYS14_25_V1
Since no one is looking at uncertainties now anyways, I just left it as a place holder

jhgoh, line 7: Ah, null string to skip JES in line 73!

jshlee, line 7: yep this is just because of PHYS14_25_V2
they should have at least filled it with zeros

PR #85, opened 2015-02-13
about keeping pass/fail and prescale of the HLTs

YoungKwonJo:
in a event:
 HLTNameList : vector string
 HLTListPassFail : vector int
 HLTPrescaleList : vector int

We can keep all triggers by single trigger set as (name, pass/fail/, prescale).
++++++++
Dataset:
vector<int>                    "recoEventInfo"             "HLTListPassFail"   "CAT"
vector<int>                    "recoEventInfo"             "HLTPrescaleList"   "CAT"
vector<string>                 "recoEventInfo"             "HLTNameList"     "CAT"
++++
However I need to check more detail.

### DIFF
--- a/CatProducer/plugins/RecoEventInfoProducer.cc
+++ b/CatProducer/plugins/RecoEventInfoProducer.cc
@@ -32,6 +32,7 @@ public:
 
 private:
   typedef std::vector<double> doubles;
+  typedef std::vector ints;
   typedef std::vector<std::string> strings;
 
   edm::EDGetTokenT<reco::VertexCollection> vertexToken_;
@@ -54,6 +55,9 @@ RecoEventInfoProducer::RecoEventInfoProducer(const edm::ParameterSet& pset)
   produces<double>("pvX");
   produces<double>("pvY");
   produces<double>("pvZ");
+  produces<strings>("HLTNameList");
+  produces<ints>("HLTListPassFail");
+  produces<ints>("HLTPrescaleList");
 
   edm::ParameterSet hltSet = pset.getParameter<edm::ParameterSet>("HLT");
   const boost::regex matchVersion("_v[0-9\\*]+$"); // regexp from HLTrigger/HLTCore/HLTConfigProvider
@@ -104,6 +108,8 @@ void RecoEventInfoProducer::produce(edm::Event& event, const edm::EventSetup& ev
   event.put(std::auto_ptr<double>(new double(pvY)), "pvY");
   event.put(std::auto_ptr<double>(new double(pvZ)), "pvZ");
 
+  strings HLTNameList;
+  ints HLTListPassFail, HLTPrescaleList;
   edm::Handle<edm::TriggerResults> hltHandle;
   event.getByToken(hltToken_, hltHandle);
   for ( auto key = hltGroup_.begin(); key != hltGroup_.end(); ++key )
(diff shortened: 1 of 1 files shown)


YoungKwonJo:
> cmsRun PAT2CAT_cfg.py useMiniAOD=True inputFiles=/store/mc/Phys14DR/DYJetsToLL_M-50_13TeV-madgraph-pythia8/MINIAODSIM/PU20bx25_PHYS14_25_V1-v1/00000/0432E62A-7A6C-E411-87BB-002590DB92A8.root

The bottom plots are the result for the new types.
vector "recoEventInfo" "HLTListPassFail" "CAT"
vector "recoEventInfo" "HLTPrescaleList" "CAT"
vector "recoEventInfo" "HLTNameList" "CAT"

However, we need to check the triggers as MuJet and ELJet. however, I didn't add the plots about it.
The values are all zero in the dataset in my test-sample.
int "recoEventInfo" "HLTELJet" "CAT" for HLT_Ele27_WP80_v
int "recoEventInfo" "HLTMuJet" "CAT" for HLT_IsoMu24_eta2p1_v

https://twiki.cern.ch/twiki/bin/viewauth/CMS/TopTrigger
The site has some information. However we need more information, recently.
+for more recent result.
https://indico.cern.ch/category/4634/
https://indico.cern.ch/event/369618/contribution/4/material/slides/1.pdf
however I can't select the triggers in the slides for using. the triggers has different in each bunch or luminosity.
What do you think? do you have a information for this?

hlt1
hlt2
hlt3

jhgoh:
Nope. I don't think it is good idea to store each names for every events.
Why don't you just add one more pset? You don't need to edit .cc.
For the prescale, I already added code skeletons, just 3 lines to activate it is enough.

CatProducer/plugins/RecoEventInfoProducer.cc
```diff
@@ -118,14 +124,23 @@ void RecoEventInfoProducer::produce(edm::Event& event, const edm::EventSetup& ev
...
+      HLTNameList.push_back(trigName);
+      const int psValue = hltConfig_.prescaleValue(event, eventSetup, trigName);
+      HLTPrescaleList.push_back(psValue);
+      
       if ( trigIndex < hltHandle->size() )
       {
-        if ( hltHandle->accept(trigIndex) ) { isPassed = true; break; }
-      }
+        //if ( hltHandle->accept(trigIndex) ) { isPassed = true; break; }
+        if ( hltHandle->accept(trigIndex) ) { isPassed = true; HLTListPassFail.push_back(1); break;}
```
jhgoh, line 134: This line have to be nested in the line 131 (originally in 121) for the safety.

``` cpp
hltHandle->accept(trigIndex);
```

sometimes fails if the trigger path name is not present in the HLTResults.

YoungKwonJo:
if you use jhgoh's solution,  we may need to separate the HLT.
for a example,
++++++

> ```
>    DoubleMu = cms.vstring(
>        "HLT_Mu17_Mu8_v*", "HLT_Mu17_TkMu8_v*","HLT_Mu17_TrkIsoVVL_TkMu8_TrkIsonVVL_v*"
>    ),
> ```
> 
> +++++
>        DoubleMu_HLT_Mu17_Mu8 = cms.vstring("HLT_Mu17_Mu8_v_"),
>        DoubleMu_HLT_Mu17_TkMu8 = cms.vstring("HLT_Mu17_TkMu8_v_"),
>        DoubleMu_HLT_Mu17_TrkIsoVVL_TkMu8_TrkIsonVVL = cms.vstring("HLT_Mu17_TrkIsoVVL_TkMu8_TrkIsonVVL_v*"),
>  +++
> as like this.

However, I have a question. can we don't care about the "v_" in end of the string?
If we don't need to care the "v_" as v1 or v2 or v3...,  I agree with you.

jshlee:
yes, separating the triggers as jhgoh's solution is fine.
Unless anyone is going to study the difference between the different versions, keeping all the versions together makes sense. 
shall we close this?

YoungKwonJo:
I agree it to close.

jhgoh:
Not merging, just close.
In addition, @YoungKwonJo you have to keep \* in the version name. Otherwise it will try to require exact matching ending with "v".

PR #86, opened 2015-02-13
bugfix-catMuonAnalysis

jshlee:


monttj:
CATMuonAnalysis should be in src area since it is single analysis source code. no?

jshlee:
It cant be, just that the CatAnalyzer/BuildFile.xml wasn't filled out and it wasn't compiling, so I moved it to plugins with the other analysis examples.
I can move it back and just add in the lines required by CatAnalyzer/BuildFile.xml

jhgoh:
plugins is correct place. src is for the libraries to be used to others.

PR #88, opened 2015-02-14
new configuration layout

jshlee:
added catTools_cff.py and patTools_cff.py
added genparticle ref to all cat objects


jshlee:
if no one has any issues with this pr, I'll merge it in 2 days

PR #89, opened 2015-02-18
updates to hlt

jshlee:
saving prescale values to trigger variable
separated all triggers
added new triggers

### DIFF
--- a/CatProducer/plugins/RecoEventInfoProducer.cc
+++ b/CatProducer/plugins/RecoEventInfoProducer.cc
@@ -111,7 +111,7 @@ void RecoEventInfoProducer::produce(edm::Event& event, const edm::EventSetup& ev
     const std::string& hltGroupName = key->first;
     const strings& hltPaths = key->second;
 
-    bool isPassed = false;
+    int psValue = 0;
     for ( auto& hltPath : hltPaths )
     {
       const strings hltPathsWithV = HLTConfigProvider::restoreVersion(hltConfig_.triggerNames(), hltPath);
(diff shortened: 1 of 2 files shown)


CatProducer/plugins/RecoEventInfoProducer.cc
```diff
@@ -120,11 +120,14 @@ void RecoEventInfoProducer::produce(edm::Event& event, const edm::EventSetup& ev
       const unsigned int trigIndex = hltConfig_.triggerIndex(trigName);
       if ( trigIndex < hltHandle->size() )
       {
-        if ( hltHandle->accept(trigIndex) ) { isPassed = true; break; }
+	if ( hltHandle->accept(trigIndex) ) {
+	  const std::pair<int,int> prescales = hltConfig_.prescaleValues(event, eventSetup, trigName);
+	  psValue = prescales.first * prescales.second;
+	  break;
```
jhgoh: Is the break; statement really needed? I think it can introduce some side effect when combined with multi-paths.
If two paths Path1(PS=10) and Path2(PS=1) are fired, it could be better to return results based on the Path2 but it will return PS factor 10.

The config in this PR splits all paths, so it is safe at this moment.

BTW, is it right to multiply prescale.first and .second? I guess one of them is just PS column number (maybe wrong)

jshlee: i didnt add the break, it was there from before.

two different prescaled triggers never be saved in the one result.

first is the L1 and second is the HLT prescales.

jhgoh: Actually I added break, not a problem without PS.
Combining 2 or more paths with PS'ed should be avoided.  
How about changing 

``` cpp
psValue = prescales.first * prescales.second;
break;
```

to 

``` cpp
psValue = min(psValue, prescales.first * prescales.second);
//break;
```

does this make sense?

jshlee:
if no one has any issues with this pr, I'll merge it in 2 days

PR #100, opened 2015-04-28
added Iso03 for muons from miniAOD

jshlee:


jshlee:
i'll merge this if there is no objections in the next few days

monttj:
Do we have any reason for this?

jshlee:
Its needed for ios 03 using miniaods.

Cheers,
Jason.

On 5 May 2015, at 6:47 pm, Tae Jeong Kim <notifications@github.com<mailto:notifications@github.com>> wrote:

Do we have any reason for this?

monttj:
Usually it is not recommended to use the isolation value from reco.
It would have been better to keep the mini-AOD variable. 
But I guess that we will have to make it configurable when producing CAT-tuple later anyway.

PR #101, opened 2015-05-13
added Jaegyun Park's extensions to electrons

jshlee:


ghost:
Ah, I got what you said, You are right.

jhgoh:
Are those variables really used in your analysis?
I don't have /strong/ objection to this but eventually it will be just a 2nd copy of existing dataformat.

jshlee:
i'll close this since i've updated it in my other pr

PR #102, opened 2015-05-13
Cat75x

jshlee:
added Iso03 for muons from miniAOD
added Jaegyun Park's extensions to electrons
added new egamma pid
added gsftrack for electrons
added PFSpecific to jets
added correctedJet() and jecFactor() to get the different correction levels for jets


jshlee:
hi all,
I would like to merge this as soon as possible.

monttj:
Do you have any estimation as to the file size per event with respect to without adding these all variables? Thank you!

jshlee:
image

jshlee:
For electrons, muons and jets, they pretty much double in size.
I think we should add in selection cuts, ie. pid, pt, eta, etc.

monttj:
It seems too much. Do we know which analysis needs these all variables? We can remove detector-based isolation variables.

monttj:
Currently, it is ~4kb per event using ttbar sample. If it is doubled, it would be around 10 kb per event. 
miniAOD is around 30-40 kb.  It will be only one third of miniAOD which is much larger than original plan.

jshlee:
yes i agree. I would like to keep it as small as possible.

jshlee:
I've removed the additional variables and the cattuples are now back to the 1/10th the miniAOD size.
If there are no objections, I'll merge this.

monttj:
no objection from my side

PR #103, opened 2015-06-09
Cat72x mini aod

jedori0228:
added SKTree variabels into catMuon.


jshlee:
i'll like to close this since it has been included in PR #102

PR #112, opened 2015-07-20
Parton top update

jhgoh:
Update PartonTopProducer

**Change particle ordering**
Top quarks are stored at the first. partons[0] and partons[1] will give top and anti-top.
No physics change is expected but there can be little modification to user analysis macros.

**Allow t->Wq, q!=b decays**
Some MC with CKM allows t->Ws t->Wd decays.
Small change in gen level event selection.

**Fix to getLast()**
Trace daughter if decays to same particle and no other decay products except photons and gluons.
In some cases, W decay products contains copy of W itself without daughter particles -
Imagine a decay chain:
   W_1(3 daughters)->W_2(0 dau) + l + nu
W_1 will be selected with this new code.
But in the decay chain
  W_a -> W_b + photons... -> l + nu + photons
W_b will be selected.


jshlee:
+1
i merged by mistake, but i dont think anyone will have any objections

PR #113, opened 2015-07-20
Flexiblility to add variable into ntuple maker

jhgoh:
Technical change, no physics change

Variables can be added with template of (generic) Consumer classes.


jhgoh:
Closing this and re open with correct remote.

PR #114, opened 2015-07-20
Jets before hadronization

jhgoh:
_Needs physics discussion with TOP-13-010 and future ttbb analyzers_

This PR adds one more jet collection from PartonTopProducer.
No change in existing objects.

In the ttbb analysis in 8TeV uses jets after parton shower before hadronization.
This definition is useful to match with theory which uses cuts.f in madgraph.

**Discussion point: input particle definition**
Select quarks or gluons which does not decay/radidate to quarks or gluons - does this give consistent definition?
Note: other particles are ignored in this jet definition.

``` c++
    else if ( absPdgId < 6 or absPdgId == 21 )
    {
      // QCD particles : select one after parton shower, before hadronization
      bool toKeep = true;
      for ( size_t j=0, m=p.numberOfDaughters(); j<m; ++j )
      {
        const int absDauId = abs(p.daughter(j)->pdgId());
        if ( absDauId < 6 or absPdgId == 21 ) { toKeep = false; break; }
      }
      if ( toKeep )
      {
        qcdParticles.push_back(&p);
      }
    }
  }
```

**Discussion point 2 : 13TeV usage**
Will this be used in 13TeV analysis?

### DIFF
--- a/CatProducer/plugins/PartonTopProducer.cc
+++ b/CatProducer/plugins/PartonTopProducer.cc
@@ -4,6 +4,9 @@
 #include "FWCore/ParameterSet/interface/ParameterSet.h"
 
 #include "DataFormats/HepMCCandidate/interface/GenParticle.h"
+#include "fastjet/JetDefinition.hh"
+#include "fastjet/ClusterSequence.hh"
+#include "RecoJets/JetProducers/interface/JetSpecific.h"
 
 #include "CommonTools/Utils/interface/PtComparator.h"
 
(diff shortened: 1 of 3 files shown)


jhgoh:
Merged by accident, reverting this soon.

YoungKwonJo:
https://twiki.cern.ch/twiki/bin/view/CMS/TTbarHbbRun2ReferenceAnalysis
https://twiki.cern.ch/twiki/bin/view/CMSPublic/GenHFHadronMatcher#Installation_instructions
I guess we need to use it or to make similar way.
The twik has the Event categorization for ttbb, ttb, tt2b, ttcc, ttlf.
Now, I am trying to add the candidates by using GenHFHadronMathcer in catTool.

However, I guess it is good point as "Discussion point: input particle definition"
However, I guess we have to match between genjet-object and b/c-hadron-candidate.

PR #115, opened 2015-07-20
Jets before hadronization

jhgoh:
Re-opening #114 

**Needs physics discussion with TOP-13-010 and future ttbb analyzers**

This PR adds one more jet collection from PartonTopProducer.
No change in existing objects.

In the ttbb analysis in 8TeV uses jets after parton shower before hadronization.
This definition is useful to match with theory which uses cuts.f in madgraph.

**Discussion point: input particle definition**
Select quarks or gluons which does not decay/radidate to quarks or gluons - does this give consistent definition?
Note: other particles are ignored in this jet definition.

``` c++
    else if ( absPdgId < 6 or absPdgId == 21 )
    {
      // QCD particles : select one after parton shower, before hadronization
      bool toKeep = true;
      for ( size_t j=0, m=p.numberOfDaughters(); j<m; ++j )
      {
        const int absDauId = abs(p.daughter(j)->pdgId());
        if ( absDauId < 6 or absPdgId == 21 ) { toKeep = false; break; }
      }
      if ( toKeep )
      {
        qcdParticles.push_back(&p);
      }
    }
  }
```

**Discussion point 2 : 13TeV usage**
Will this be used in 13TeV analysis?

### DIFF
--- a/CatProducer/plugins/PartonTopProducer.cc
+++ b/CatProducer/plugins/PartonTopProducer.cc
@@ -4,6 +4,9 @@
 #include "FWCore/ParameterSet/interface/ParameterSet.h"
 
 #include "DataFormats/HepMCCandidate/interface/GenParticle.h"
+#include "fastjet/JetDefinition.hh"
+#include "fastjet/ClusterSequence.hh"
+#include "RecoJets/JetProducers/interface/JetSpecific.h"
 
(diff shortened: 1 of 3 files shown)


jhgoh:
Closing this and re open with correct remote.

PR #116, opened 2015-07-20
Flexiblility to add variable into ntuple maker

jhgoh:
#113

Re-open with correct remote.

Technical change, no physics change

Variables can be added with template of (generic) Consumer classes.


jshlee:
:+1:

PR #118, opened 2015-07-21
add catGetHFHadronMatching_cff

YoungKwonJo:
https://twiki.cern.ch/twiki/bin/view/CMS/TTbarHbbRun2ReferenceAnalysis#TT_X_splitting
https://twiki.cern.ch/twiki/bin/view/CMSPublic/GenHFHadronMatcher#Output_of_the_tool

### DIFF
--- a/CatProducer/prod/PAT2CAT_cfg.py
+++ b/CatProducer/prod/PAT2CAT_cfg.py
@@ -36,6 +36,9 @@ patTool(process, runOnMC, useMiniAOD)
 from CATTools.CatProducer.catTools_cff import *
 catTool(process, runOnMC, doSecVertex, useMiniAOD)
 
+from CATTools.CatProducer.catGetHFHadronMatching_cff import *
+genHFTool(process, runOnMC,  useMiniAOD)
+
 from CATTools.CatProducer.catEventContent_cff import *
 process.out.outputCommands = catEventContent
 if runOnMC:
--- a/CatProducer/prod/PAT2CAT_cfg.py
+++ b/CatProducer/prod/PAT2CAT_cfg.py
@@ -36,8 +36,8 @@ patTool(process, runOnMC, useMiniAOD)
 from CATTools.CatProducer.catTools_cff import *
 catTool(process, runOnMC, doSecVertex, useMiniAOD)
 
-from CATTools.CatProducer.catGetHFHadronMatching_cff import *
-genHFTool(process, runOnMC,  useMiniAOD)
+from CATTools.CatProducer.catGenHFHadronMatching_cff import *
+genHFTool(process, useMiniAOD)
 
 from CATTools.CatProducer.catEventContent_cff import *
 process.out.outputCommands = catEventContent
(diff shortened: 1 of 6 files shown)


CatProducer/prod/PAT2CAT_cfg.py
```diff
@@ -36,6 +36,9 @@
 from CATTools.CatProducer.catTools_cff import *
 catTool(process, runOnMC, doSecVertex, useMiniAOD)
 
+from CATTools.CatProducer.catGetHFHadronMatching_cff import *
+genHFTool(process, runOnMC,  useMiniAOD)
```
jshlee: a bit pedantic, but this shouldnt need runOnMC as input since it should be only used for mc

jshlee:
Hi YoungKwon,

Is there a nicer way to put this in to catJets.
i.e. catJets already have;
    int partonFlavour_;
    int hadronFlavour_;
    int partonPdgId_;
https://github.com/vallot/CATTools/blob/cat75x/DataFormats/interface/Jet.h#L101-L104

YoungKwonJo:
ttbb categorization,
  ttbb : 2 bjet with each matched one b-hadron for additional jet(not from top)
  ttb   : 1 bjet with matched one b-hadron  for additional jet(not from top)
  tt2b : 1 bjet with matched two b-hadron for additional jet(not from top)
  ttcc  : 2 cjet with matched c-hadron for additional jet(not from top)
  ttlf    : 2 lfjet with light flavor for additional jet(not from top)

We have to define the categorization for TTJets.
However, by using the values  in catJets, we easily can't make the categorization.
We need to define more values for keeping all-variables.

jshlee:
i see what u mean

YoungKwonJo:
https://twiki.cern.ch/twiki/bin/view/CMSPublic/GenHFHadronMatcher#Event_categorization_example_1
Perhaps we can reduce the variables by using the guide-line in the link.
https://twiki.cern.ch/twiki/pub/CMSPublic/GenHFHadronMatcher/GenTtbarCategorizer.cc
https://twiki.cern.ch/twiki/pub/CMSPublic/GenHFHadronMatcher/GenTtbarCategorizer_cfi.py.txt

However, I can't find the code is not in cmssw-git.
However, if we make a solution by using the links, we can define only one-value for the categorization.

I will trying to make the solution.

YoungKwonJo:
I added it as the GenTtbarCategorizer in CatProducer.
int                            "GenTtbarCategories"        "genTtbarId"      "CAT"
we will keep only one variable in edm.

monttj:
GenTtbarCategorizer.cc is already available in CMSSW.
Is there any reason why we implemented the same in CAT area?

YoungKwonJo:
No, GenTtbarCategorizer.cc is not in CMSSW.
https://twiki.cern.ch/twiki/bin/view/CMSPublic/GenHFHadronMatcher#Event_categorization_example_1
I used the code in the twiki.

The GenHFHadronMatcher.cc is already in CMSSW.
However, the output has too many branchs of GenHFHadronMatcher as the link.
https://github.com/cms-sw/cmssw/blob/CMSSW_7_6_X/PhysicsTools/JetMCAlgos/plugins/GenHFHadronMatcher.cc#L150

So, if we use the GenTtbarCategorizer, we can make one-value as the link.
https://github.com/vallot/CATTools/blob/cat75x/CatProducer/plugins/GenTtbarCategorizer.cc#L133

PR #119, opened 2015-07-21
add a variable in edm by using catGetHFHadronMatching_cff

YoungKwonJo:
https://twiki.cern.ch/twiki/bin/view/CMS/TTbarHbbRun2ReferenceAnalysis#TT_X_splitting
https://twiki.cern.ch/twiki/bin/view/CMSPublic/GenHFHadronMatcher#Event_categorization_example_1

add 
python/GenTtbarCategorizer_cfi.py
plugins/GenTtbarCategorizer.cc

### DIFF
--- a/CatProducer/plugins/GenTtbarCategorizer.cc
+++ b/CatProducer/plugins/GenTtbarCategorizer.cc
@@ -1,0 +1,375 @@
+// -*- C++ -*-
+//
+// Package:    PhysicsTools/GenTtbarCategorizer
+// Class:      GenTtbarCategorizer
+// 
+/**\class GenTtbarCategorizer GenTtbarCategorizer.cc PhysicsTools/JetMCAlgos/plugins/GenTtbarCategorizer.cc
+//https://twiki.cern.ch/twiki/pub/CMSPublic/GenHFHadronMatcher/GenTtbarCategorizer.cc
+//
+ Description: Categorization of different tt+xx processes, returning unique ID for each process as e.g. tt+bb, tt+b, tt+2b, tt+cc, ...
+
+ Implementation:
+     
+     The classification scheme returns an ID per event, and works as follows:
+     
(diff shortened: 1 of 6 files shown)


jshlee:
hi @YoungKwonJo

Shall we merge this?

YoungKwonJo:
Yes, I would like to merge it. 
If we have no problem about merge.

jhgoh:
Questions:
- Doesn't it change the event content of catTuple and event size?
- Isn't it already in CMSSW or planned to be included there?

YoungKwonJo:
1.  we will keep only one-value in catTuple.
   +++++int "GenTtbarCategories" "genTtbarId" "CAT"+++++
2. I have tested it in CMSSW CMSSW_7_4_6_patch6. we don't need to install(download) any-code in the cmssw version.
   However, in old version, we need to add some code. the following is the instruction. 
   https://twiki.cern.ch/twiki/bin/view/CMSPublic/GenHFHadronMatcher#Installation_instructions

jshlee:
we are using CMSSW_7_4_6_patch6 so it should be fine
so this only saves one int per event?

YoungKwonJo:
Yes, it saves one int per event.
the candidates are followings.
+++++++++++++++++
(x can take values 0, 1, 2 depending on the number of b jets from top):
x51: tt+b 
x52: tt+2b 
x53, x54, x55: tt+bb 
x41, x42, x43, x44, x45: tt+cc 
x00: tt+LF
+++++++++
https://twiki.cern.ch/twiki/bin/view/CMSPublic/GenHFHadronMatcher#Event_categorization_example_1

And,  I tested it by using only MINIAODSIM. 
Do you think to test by using AODSIM?

jshlee:
for aodsim, do u want to talk on skype?

YoungKwonJo:
sent a message in your mail.

PR #123, opened 2015-08-04
updated JEC and added puppi jet jec correction

jshlee:


monttj:
Can we have the name like "CATVertex" "CATTrigger" instead of "CATVertexInfo" and "CATTriggerInfo" so that it can be consistent with "CATMuon" or "CATJet"?

jshlee:
yep sure

PR #125, opened 2015-08-05
Temporary simple PU weight from reco vertex

jhgoh:
A simple PU weight method is added based on the reco vertex size.
This method is non-standard and not recommended for long term.

There is a handle to turn on/off this non-standard method, "isStandardWeight = True" will move it back to standard method.
Weight values are taken from the Phys14 powheg ttbar sample and Run2015B data, contribution by Youn.


jhgoh:
No runtime checks done. I hope automatic tool to be prepared sometime.

PR #126, opened 2015-08-05
added pseudoTop

jshlee:
since it cannot be run from cat


jhgoh:
We need followings to the production recipe

```
cd $CMSSW_BASE/src
git-cms-addpkg TopQuarkAnalysis/TopEventProducers
git-cms-merge-topic jhgoh:PseudoTop
```

To be synchronised with the TOP RIVET version.

jshlee:
yep.
so we should do this because it is needed right?

jhgoh:
Yes.

PR #127, opened 2015-08-05
added catvertex and cattrigger

jshlee:


jhgoh:
Why don't you just use reco::Vertex?

jshlee:
size

jhgoh:
I meant saving new reco::VertexCollection with only one object without associating tracks. Information must be same as cat::Vertex and we also have compatibility to standard CMSSW modules.

PR #131, opened 2015-08-05
changed to reco vertex without tracks

jshlee:


jshlee:
this is most likely to be the prod version

Issue #53, opened 2015-01-22
dataset management


jshlee:
using either JSON or xml to bind dataset/reference, globaltags, cfg/cross section info, number of events

be able to be read by analysis and crab submitting scripts /and displayed on web if possible

geonmo:
To support Xsec information is required next release.

jshlee:
can you document this in the twiki?
or we can add in a file list in the analyser dir

jhgoh:
One possible solution is to save sample info in google spreadsheet and export to csv format.

https://docs.google.com/spreadsheets/d/1rWM3AlFKO8IJVaeoQkWZYWwSvicQ1QCXYSzH74QyZqE/edit?usp=sharing
- https://docs.google.com/spreadsheets/d/1rWM3AlFKO8IJVaeoQkWZYWwSvicQ1QCXYSzH74QyZqE/pub?gid=0&single=true&output=csv
- https://docs.google.com/spreadsheets/d/1rWM3AlFKO8IJVaeoQkWZYWwSvicQ1QCXYSzH74QyZqE/pub?gid=1289302456&single=true&output=csv

These csv files can be parsed by a python script to generate fileList + cross section JSON file then passed to user analysis macros.

jhgoh:
"makeFileList" is updated (with new name getDatasetInfo) to retrieve dataset information from the google doc.

jhgoh:
getDatasetInfo is updated. It produces simple txt formatted filelist to be used from create-batch or other tools.
JSON output can be used from the python scripts for plotting.

jhgoh:
I think this issue can be closed since we have google doc -> CSV -> JSON interface.

PR #136, opened 2015-08-05
New script to dump file list

jhgoh:
Added new script to make file list in JSON format.

Run the script with all LFN's in the argument or type them interactively to create samples.json file.
Cross section values are not considered in this output file.


jhgoh:
quoting this PR for tracking. #53

PR #156, opened 2015-08-14
Update README.md

geonmo:
Change to "git" protocol to "https" protocol for users which did not setup a public key.


jshlee:
all users should have really setup up a public key

geonmo:
@jshlee Yeah, but it is just recommend.

PR #160, opened 2015-08-16
switch to stream EDProducer, unique_ptr

jhgoh:
switch to stream::EDProducer
use std::unique_ptr


jshlee:
shall we add this for the new prod?

jhgoh:
After doing runtime test.

jshlee:
are u going to test?

jhgoh:
Did you update cattools install. recipe - electron id git-cms-merge-topic?

PR #164, opened 2015-08-20
fix for v7-3-6

jshlee:
fix memory leak in jet producer
added all ele and mu triggers as vector


jhgoh:
Do you have evt size estimation for catTrigger?

jshlee:
Stringintstdpairs_catTrigger__CAT. 411.79 78.08

jshlee:
I only save the ele and mu triggers that pass

jshlee:
i put this in because the single muon trigger wasnt working

PR #169, opened 2015-08-25
updated electron ID to cutBasedElectronID-Spring15-50ns-V1

jshlee:


jshlee:
shall i start a new production with this update?

jshlee:
will merge this since it will be in new prod.

PR #183, opened 2015-09-08
For 25ns production

jshlee:
shall we have a new branch for 25ns
or have a flag to switch from 50 to 25ns?
currently, I just overwrote all 50 settings with 25ns.


jhgoh:
Having new branch seem to be reasonable.
Analysis tool codes should be back ported to 50ns branch as well.

jhgoh:
Are we using the same release 747p2? Are you going to update Readme.md as well?

geonmo:
@jhgoh  I think so. It is better to separate branchs for 25 and 50ns.

jshlee:
will merge and start 25ns production!

jshlee:
@jhgoh yes we will start with 747p2
@geonmo - I'll split into new branch or have flags to change between 25 and 50 settings after having a meeting. 
for now I will just tag this as v7-4-0

Issue #163, opened 2015-08-18
Kinematic reconstruction


jhgoh:
We got codes from DESY group. Consider to port them into CMSSW or CATTools.

jhgoh:
Still in progress to port DESY code into CATTools.

jhgoh:
Some discussion points gone to the PR #184.
New KinReco code uses M(l,b) for the weighting factor (or M(ttbar)). This is largest change from the TQAF version.

jhgoh:
@kuyoun , please post input/output p4 from DESY here.

jhgoh:
Event list from youn:

```
event: 284322939 
lep1 pt: 20.9255 lep1 eta: 0.617559 lep1 phi: 2.36695 lep1 m: -0.00560447 
lep2 pt: 21.5264 lep2 eta: 0.481956 lep2 phi: 1.06914 lep2 m: 0.1057
 event: 284322939  0th jet pt: 46.2721 jet eta: -1.72551 jet phi: -1.68423 jet m: 8.45819
 event: 284322939  1th jet pt: 45.9963 jet eta: 1.26061 jet phi: -0.810852 jet m: 7.42647
 event: 284322939  2th jet pt: 43.675 jet eta: 0.925888 jet phi: -2.26665 jet m: 7.83945
No-hf-met => pt: 82.3015 phi: 1.07359
Top quark => pt: 72.4154 eta: 2.16134 phi:  2.05428
Antitop quark => pt: 51.5434 eta: -1.7133 phi:  -0.822213
==========================================================================

 event: 286791895 
lep1 pt: 77.4678 lep1 eta: 0.520264 lep1 phi: -1.18907 lep1 m: 0.016377 
lep2 pt: 56.5712 lep2 eta: -0.104898 lep2 phi: 2.32963 lep2 m: 0.1057
 event: 286791895  0th jet pt: 67.3577 jet eta: -0.610829 jet phi: -2.58806 jet m: 8.97971
 event: 286791895  1th jet pt: 43.0293 jet eta: -0.271075 jet phi: 0.47781 jet m: 11.3372
No-hf-met => pt: 90.4005 phi: 0.202267
Top quark => pt: 151.88 eta: 0.871251 phi:  -0.567418
Antitop quark => pt: 86.9487 eta: -0.433286 phi:  2.48138
==========================================================================

 event: 289496079 
[...]

PR #184, opened 2015-09-08
Kinematic solver from DESY

jhgoh:
Kinematic solver codes from DESY is going to be ported to CATTools. #163


jhgoh:
Some updates are to be added.

jhgoh:
Users can use new Kinematic solver from DESY group within CATTools. We need some validation/synchronization to the reference.

Definition of weight is completely changed - M(l,b) distribution from the MC truth is used to extract weight, rather than the neutrino energy flux. 1/M(ttbar) is used for the "mass loop" algorithm but this is not the default weighting method.

jhgoh:
Currently, this code set assumes CMSSW framework, but there are some request to have similar implementation only with bare root. I would like to have another PR to do refactoring of the codes.

PR #185, opened 2015-09-09
Update recipe to include backported NewGenHFHadronMatcher

jhgoh:
Update recipe to include backported NewGenHFHadronMatcher


jhgoh:
GenHFHadronMatcher is updated to correctly work on charmed particles.
This update is available in 76X but no official backport yet, thus YoungKwon prepared a branch to include it.

PR #187, opened 2015-09-09
changed to FrontierConditions_GlobalTag_condDBv2_cff

jshlee:


jshlee:
i think this is the last commit before production
i've tested this with 25ns mc and rd and everything looks good so far

PR #191, opened 2015-09-10
Split MC top workflow

jhgoh:
Split workflow for MC top quarks.
Produce pseudoTop, partonTop only if runGenTop=True
Within the submitCrab3.py, this option is automatically turned on if the dataset name starts with /TT (or /tt).


jhgoh:
This PR is not urgent, but this code changes can be useful to turn on/off the legacy catGenTop object.

PR #195, opened 2015-09-15
CATTriggerPacker for new trigger dataformat

jhgoh:
CATTriggerPacker is updated to use edm::TriggerResults and patTrigger for the prescale factors.
Some options are changed, to choose combine by or/and.

Note : There are some code duplication with CATTriggerProducer.


jshlee:
:+1:

jhgoh:
This PR is to be merged AFTER the next production because of dependency to the catTuple outputContent definition.

PR #200, opened 2015-09-16
Separate KinReco module

jhgoh:
#163

Separate candidate combination and kinematic reco module.

Ported version of KinReco seem to have some problem to reproduce top mass, this have to be fixed.


jhgoh:
For a tracking - mTop1, mTop2, mW1 and mW2 plots from various KinReco algorithms.
I suspect some bug in TopKinRecoUtils.cc... need investigation.

cmt1
cmt2
cmw1
cmw2

jhgoh:
Bug found. fix will be provided soon.

jhgoh:
m1
m2

jhgoh:
m1
m2
m3
m4

jshlee:
what changed??

jhgoh:
There were typos in the routine to extract coefficients, thus top, W mass from the fit was wrong. The last commit fixes the problem thus gives correct top and W masses.

jhgoh:
Fixed conflicts
Use random number generator service
Fixed python syntax error

PR #212, opened 2015-09-25
Run on relval sample

jhgoh:
The previous attempt to run on relval was affecting the central production - can be solved by adding an option to control to run on metSource hack to run on relval sample.


jhgoh:
On hold, need some idea to run on relval without option. It is very important not to affect the central production. Previous attempt was a blocker since it replaces metSource during python parsing but crab does not know this must be reverted to run on grid.

PR #214, opened 2015-09-25
Pseudo top analysis

jhgoh:
CATGenLevelAnalysis.cc is updated to produce parton vs particle level top comparison plots.
Binning is taken from TOP-12-028, which is going to be adopted in TOP ModGen PAS (in preparation).
The scale factors (or correction functions) and response matrices parton -> particle level are going to be used in the PAS mentioned above.

Lepton, b quark(bjet) distributions are omitted in this PR to concentrate on the variables in agreement in TopModGen group but the extension is trivial.


jhgoh:
On hold to add some macro files.

jhgoh:
Releasing since the scripts are added and plots can be produced without crash.

PR #220, opened 2015-09-28
Generator level dilepton analysis

jhgoh:
An example generator level dilepton analysis code is added.
The analysis code produces generator level dilepton mass, pt, eta, phi and other useful distributions with find binning. (1GeV for mass and pt, 0.05 radian for phi, 0.01 for eta)

The generator level leptons are basically parton level objects - particles before radiation.

First usage of this analyzer is of course quick comparison in generator level.
But the main motivation to have this analyzer is calculating mass filter efficiency ( 20 < M < 50 ) in low mass Drell-Yan sample. The value will be then used to extrapolate cross section at > 20GeV to >10GeV samples.

### DIFF
--- a/CatAnalyzer/data/.gitignore
+++ b/CatAnalyzer/data/.gitignore
@@ -1,0 +1,2 @@
+dataset.json
+dataset_*.txt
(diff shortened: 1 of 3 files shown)


jhgoh:
Result with DY10-50 catTuple:
screenshot

Here, the ratio of 20<M<50 to 10<10<50 is 0.200123.

```
cout << hZMass->Integral(20,50)/hZMass->Integral(10,50) << endl;
0.200123
```

jhgoh:
To be merged after double checking with normal DYJets sample.

jhgoh:
canvas_1

DYJets mass distribution is obtained using this module.
Cross section is obtained using the ratio between 20<M<50 and 10<M<50, but the mass distribution is not smooth thus I added arbitrary factor 1.14 to match ZJets and low mass DY sample (in the google spreadsheet).

One more: Neutrino energy spectrum 2D plot is added for future ttbar kinematics study.

PR #230, opened 2015-09-30
More plots GenLevel analyzer

jhgoh:
Added more plots in the CATGenLevelAnalysis module.
Now 2D plots to see migrations between channels are available.

plus acceptance cut is changed in single lepton channel.


jhgoh:
Work in progress, to be merged after finishing fine tunes.

jhgoh:
To be merged after #232

PR #231, opened 2015-10-01
revert back to old muon relIso

jshlee:


jshlee:
Hi @monttj, did you want electrons to be done the same way too?

monttj:
@jshlee 
Thank you for the update. Yes, it would be great if we can have a consistent setup for electrons.

jshlee:
but for electrons, relIso needs both rhoIso and AEff, which is different to muons.
however, both electrons and muons both have the same absIso(float dR=0.3, float dBetaFactor=0.5) function. so, relIso can both be calculated the same way with absIso/pt.

jshlee:
shall i merge this for now and think about electrons later on???

jhgoh:
+1 do the rest with separate PR.

monttj:
Hello,

I think there is no reason why we should not implement the dbeta correction.
Currently this is only method we can rely on to reduce the pileup effect.
Without it, we can not use the electron properly. So please add dbeta for electron.
We can think about the rho later. The effective area should also be given by the POG group.

PR #235, opened 2015-10-01
Unit tests for sample productions

jhgoh:
More unit tests are added, with and without generator level top stuffs.


jhgoh:
This PR changes PF2CAT.cfg which is used for the production.
But the change is trivial and never affects actual production workflow. run test file is independent from the production.
Safe to be merged.

PR #237, opened 2015-10-04
More unit tests for production

jhgoh:
Add unit tests using the real data, use inputFiles option.


jhgoh:
This PR is only for the unit test, no effect on the central production. Merging this PR myself.

PR #242, opened 2015-10-06
pseudoTop channel modified

tt8888tt:
channel only for now. mode will be added.

### DIFF
--- a/CatAnalyzer/plugins/TtbarDiLeptonAnalyzer.cc
+++ b/CatAnalyzer/plugins/TtbarDiLeptonAnalyzer.cc
@@ -62,7 +62,9 @@ private:
 
   TTree * ttree_;
   int b_partonChannel, b_partonMode1, b_partonMode2;
-  int b_pseudoTopChannel, b_pseudoTopMode1, b_pseudoTopMode2;
+  int b_pseudoTopChannel;
+  float b_pseudoToplep1_pt, b_pseudoToplep1_eta;
+  float b_pseudoToplep2_pt, b_pseudoToplep2_eta;
   int b_njet, b_nbjet, b_step, b_channel;
   bool b_lepinPhase, b_jetinPhase;
   float b_MET, b_maxweight;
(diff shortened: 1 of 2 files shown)


jshlee:
shall we merge when the mode is added?

tt8888tt:
if we don't need pseudoTop_mode I think it's good to be merged now

CatAnalyzer/plugins/TtbarDiLeptonAnalyzer.cc
```diff
@@ -318,9 +310,11 @@ void TtbarDiLeptonAnalyzer::analyze(const edm::Event& iEvent, const edm::EventSe
   if      ( b_channel == CH_ELEL ) iEvent.getByToken(trigTokenELEL_, trigHandle);
   else if ( b_channel == CH_MUMU ) iEvent.getByToken(trigTokenMUMU_, trigHandle);
   else if ( b_channel == CH_MUEL ) iEvent.getByToken(trigTokenMUEL_, trigHandle);
-  if ( !trigHandle.isValid() ) return;
+  if ( !trigHandle.isValid() ){
+    ttree_->Fill();
+	return;
+  }
   b_tri = *trigHandle;
-  //if (!tri) return;
```
jhgoh, line 323: You can modify this part something like

``` c++
if ( trigHandle.isValid() ) b_tri = *trigHandle;
```

to continue to next step even if the module fails to get the trigger bit info.

jshlee, line 323: does the trigger bit fail?

jhgoh, line 323: No if cfg is written correctly.

jshlee, line 323: if ( !trigHandle.isValid() ){cout <<"error, trigger not vaild "<<endl;} 

might be easier to debug

tt8888tt, line 323: I just checked 'channel != 0 && tri == -9' cut returns 0 events for every mc&data with this code. so I think we can get rid of 'trigHandle.isValid()' test.(?)

jshlee, line 323: i think it will still be useful to debug. we can add this to the unit test too.

jhgoh, line 323: Personally I prefer to remove isValid() part to raise exception.

CatAnalyzer/plugins/TtbarDiLeptonAnalyzer.cc
```diff
@@ -318,9 +310,8 @@ void TtbarDiLeptonAnalyzer::analyze(const edm::Event& iEvent, const edm::EventSe
   if      ( b_channel == CH_ELEL ) iEvent.getByToken(trigTokenELEL_, trigHandle);
   else if ( b_channel == CH_MUMU ) iEvent.getByToken(trigTokenMUMU_, trigHandle);
   else if ( b_channel == CH_MUEL ) iEvent.getByToken(trigTokenMUEL_, trigHandle);
-  if ( !trigHandle.isValid() ) return;
+  if ( !trigHandle.isValid() ){ cout <<"error, trigger not vaild "<<endl; }
   b_tri = *trigHandle;
```
jhgoh, line 313: Actually you don't need to printout this since the line 314 will raise exception automatically and the automatic message contains what which InputTag was asked etc.

jshlee, line 313: yep thats fine with me.
just remove line 313

PR #243, opened 2015-10-06
choose min mTTbar among four solutions for smeared solver

jhgoh:
Solver with smearing is updated to select mininum mTTbar among four solutions.


jhgoh:
#163

jhgoh:
This PR does not affect any of currently active analyses. Safe to be merged.

PR #244, opened 2015-10-07
updated relIso in electrons

jshlee:
to be similar to muons when used with dBetaFactor as requested by @monttj


DataFormats/interface/Electron.h
```diff
@@ -60,6 +56,14 @@ namespace cat {
       float charged = chargedHadronIso(dR);
       return charged + ( corNeutralIso>0 ? corNeutralIso : 0 ) ;
     }
+    float relIso(float dR=0.3, float dBetaFactor = 0) const {
+      if (dBetaFactor > 0){
```
jhgoh: dBetaFactor >= 0 would be better.

jshlee:
done!

PR #245, opened 2015-10-08
Shift channel number

jhgoh:
Channel & decay mode number from PartonTopProducer is shifted by -1.
CH_NONE is now assigned to -1, so valid channel/mode numbers start from zero.

@kuyoun , @tt8888tt, @dhkim81 The TtbarDileptonAnalyzer module reads this information but it just dumps to the ntuple. If your macro is relying on this information in your macro already and don't feel to switch to new numbering scheme, I'll withdraw this new numbering scheme.


jshlee:
what about keeping the channel numbers the same and just have CH_NONE = -1?
or having a common enum defined somewhere.

jhgoh:
With this PR and the most recent version of PseudoTopProducer, 
PartonTop and PseudoTop have same numbering scheme to set CH_NONE=-1.

I'm merging this to keep consistency between two producers before the central production.

Issue #189, opened 2015-09-10
Pileup reweighting


jhgoh:
Current PU re-weighting in CATTools is based on number of offline PV which is not standard one.
We are waiting for the central PU JSON and MinBias cross section & syst. uncertainty variation factors.

The true PU distributions are available in the CMSSW
- SimGeneral/MixingModule/python/mix_2015_50ns_Startup_PoissonOOTPU_cfi.py
- SimGeneral/MixingModule/python/mix_2015_25ns_Startup_PoissonOOTPU_cfi.py

There is a presentation in HWW group (Sep 3)
- https://indico.cern.ch/event/439320/session/2/contribution/28/attachments/1149023/1648504/Sep-03_HWW_RunII_Massironi.pdf
  This is based on private PU JSON by Mike and an extrapolation xsec~70000ub from P. Skands.

jhgoh:
(info from TOP xsec meeting) No official recipe yet. Still waiting.

jshlee:
i think this can be closed since we are upto date with it.

PR #249, opened 2015-10-12
2015 Preliminary Pileup weight

jhgoh:
New but preliminary recipe is available here
https://hypernews.cern.ch/HyperNews/CMS/get/physics-validation/2451/1/2/1/1/2/1.html

80mb +- 5% uncertainty for pileupCalc.py input parameters, MC pu distribution from CMSSW startup 25ns cfg.

There was a bug while running accumulate function, initial value "0" had to be "0.0".


jhgoh:
#189

jshlee:
since we dont really need to save the 3 variables because they are all derived from 1 variable.
how about we make this similar to the JER?
something that can be called quickly from FWlite.

jhgoh:
Basically no, otherwise we have to 
- store addPileupInfo as whole 
- or re-implement PU reweighting algorithms from the scratch. 

I think saving 3 numbers is not harmful.

jshlee:
ah because of the other bx..
no worries then

jshlee:
merge?

jhgoh:
Can anybody check weight distribution using the official sample? I tried to run it using the relval sample but this just gives only one weight & nTrueInteraction.

jshlee:
we need some testers!

jshlee:
screen shot 2015-10-13 at 8 59 40 pm
screen shot 2015-10-13 at 8 59 32 pm
screen shot 2015-10-13 at 8 59 22 pm

there is a lot of events with weight of 0. is this normal?

jhgoh:
Bugfix to be added soon.

jhgoh:
Output from the pileupCalc.py, "true" mode, mininum bias cross section = 80mb +- 5%.
Peak position is at ~10.

screen shot 2015-10-14 at 12 08 44 pm

Printing out the nTrueInteractions from the PileupInfo, it gives 35 and does not change in event.

jhgoh:
Comparison between true and observed mode
screen shot 2015-10-14 at 12 16 05 pm

jshlee:
merge?

jhgoh:
Still in doubt to use this out of box, but let us merge now and look at the distr. with full stat.

PR #250, opened 2015-10-13
added btagging sf

jshlee:


DataFormats/src/Jet.cc
```diff
@@ -64,3 +64,108 @@ float Jet::smearedRes(int direction) const {
   const double fJER = std::max(0., (genJetPt+dPt*cJER)/jetPt);
   return fJER;
 }
+
+float Jet::scaleFactorCSVv2(int op, int sys, int flav) const {
+  if (std::abs(this->pt()) > 2.4 ) return -1; // reject jets out of eta range
```
jhgoh: this->pt() should be this->eta()

DataFormats/src/Jet.cc
```diff
@@ -64,3 +64,108 @@ float Jet::smearedRes(int direction) const {
   const double fJER = std::max(0., (genJetPt+dPt*cJER)/jetPt);
   return fJER;
 }
+
+float Jet::scaleFactorCSVv2(int op, int sys, int flav) const {
+  if (std::abs(this->eta()) > 2.4 ) return -1; // reject jets out of eta range
+  //based on https://twiki.cern.ch/twiki/bin/viewauth/CMS/BtagRecommendation74X50ns
+  if (op == 0 && sys == 0 && flav == 1 && this->pt() > 30 && this->pt() < 670 ) return 0.901434*((1.+(0.0852659*this->pt()))/(1.+(0.0769021*this->pt())));
+  if (op == 0 && sys == 0 && flav == 0 && this->pt() > 30 && this->pt() < 670 ) return 0.901434*((1.+(0.0852659*this->pt()))/(1.+(0.0769021*this->pt())));
```
jhgoh: These two lines give identical formula. Is this correct?

jshlee: yes it is.
same for medium and tight

jhgoh: And what happens if op=0 && flav=anything but very high pT jets?
Logically this function will return -1 but this is un natural to me. Usual choice must be simply extrapolating SF to higher pT or just use SF=1.

jshlee: i dont think users should be using anything higher that 670 gev.

jshlee: or we can return 1 with +- error 100%

jhgoh:
@jshlee I sent a PR to your repo.

jshlee:
will merge this
can be updated laters if required.

PR #256, opened 2015-10-19
Color Coherence code

hyunyong:


jshlee:
add in the cfg file too

PR #260, opened 2015-10-20
Use r-value reference syntax

jhgoh:
Use the r-value reference syntax in C++11 standard - better readability but no cost of un-necessary copy constructors.

Purely technical update. No change in physics.


jshlee:
merge?

jhgoh:
SVP

PR #270, opened 2015-10-27
TTdilepDraw clean up

tt8888tt:
applied reading google doc system, and simplized few part.
(same 'TTdilepDraw clean up' commit before, just re-reqeust it, to avoid conflicts.)


jhgoh:
no strong objection on this, this is analysis level plotting macro.

PR #277, opened 2015-11-03
PU weight on the fly

jhgoh:
PUWeight can be applied onto the CATTuples.
User can select reweighting mode "Standard", "OnTheFly" or "NVertex".

Standard : Read AddPileupInfo to calculate PU weights
Redoweight : Read int type nTruePileup - use one of the output of the CATTuple
NVertex : Bin-by-bin correction on the number of vertex - only for (too) early analysis


jhgoh:
Test is done. it runs without crash (test suite in CatProducer/test/testRerunPUWeight_cfg.py).

PR #282, opened 2015-11-04
added root colours to catGetDatasetInfo

jshlee:


CatAnalyzer/scripts/catGetDatasetInfo
```diff
@@ -44,6 +45,7 @@ for l in csv[1:]:
                'xsec':float(xsec),
                'nevt':int(nevt),
                'lumi':float(lumi),
+               'colour':int(colour),
```
jhgoh, line 48: In the MakeHist.py, this can be improved with eval() statement. By this, 'kRed+1' can be addressed.

jshlee, line 48: i think ints are more flexible even thou they are not easy to translate for humans

jshlee:
will merge for now since students are quickly going to use this to plot results for tmr

PR #283, opened 2015-11-04
Met (jet) systematics

jalmond:


jhgoh:
This introduces dataformat change. Is there any plan for new (or re)prod?

jshlee:
yep
so that we can include silver json
john was testing the met uncertainties and found the jet ones doesnt agree with the jes and jer from jets

PR #286, opened 2015-11-04
Color coherece update 04Nov

hyunyong:


CatAnalyzer/test/run_ColorCoherence_b_cfg.py
```diff
@@ -0,0 +1,54 @@
...
+
+process.load("FWCore.MessageService.MessageLogger_cfi")
+process.maxEvents = cms.untracked.PSet( input = cms.untracked.int32(-1) )
+
+process.source = cms.Source("PoolSource",
+    fileNames = cms.untracked.vstring()
+)
+in_txt = open(sys.argv[2])
+for f in in_txt:
+    process.source.fileNames.append(f)
```
jhgoh: Just for a question - does this work? I usually do this with

``` python
for f in in_txt.readlines():
    process.source.fileNames.append(f.strip())
```

jshlee: should work, but i think HY should remove this and use catGetDatasetInfo and create-batch

PR #296, opened 2015-11-09
made submitCrab3 and checkDataSets friendly with catGetDatasetInfo

jshlee:


CatAnalyzer/scripts/catGetDatasetInfo
```diff
@@ -44,6 +44,9 @@ nevtIdx = csv[0].index("NEvent")
 lumiIdx = csv[0].index("luminosity (pb-1)")
 colourIdx = csv[0].index("colour")
 pathIdx = csv[0].index("Path")
+DataSetNameIdx = csv[0].index("DataSetName")
+GlobalTagIdx = csv[0].index("GlobalTag")
+LumiMaskIdx = csv[0].index("LumiMask")
```
jhgoh, line 49: Do we really need this?

CatAnalyzer/scripts/catGetDatasetInfo
```diff
@@ -56,7 +59,24 @@ for l in csv[1:]:
...
+
+import json
+f = open("%s/dataset.json" % outDir, "w")
+print "Saving dataset info to %s/dataset.json" % outDir
+print>>f, json.dumps(ds)
+f.close()
+
+# since all files are at kisti
+if not "sdfarm" in hostname:    
+    sys.exit()
```
jhgoh, line 79: I think L78-79 is not necessary. Files are only at kisti, but the script gives other infos for styling, cross section, etc.

CatAnalyzer/scripts/catGetDatasetInfo
jshlee, line 49: I thought it would be easier to keep all dataset info in one place.
and have crab submission read from that too.

CatAnalyzer/scripts/catGetDatasetInfo
jshlee, line 79: yep so at kisti, all files are made and at other places only the dataset.json

CatAnalyzer/scripts/catGetDatasetInfo
jhgoh, line 49: Ah OK this is for the production.

PR #300, opened 2015-11-12
catGetDatasetInfo lists prod directories too

jshlee:
have catGetDatasetInfo check if database is correct while creating sample list files


jshlee:
updated to pick the latest version of the production

jhgoh:
Hmm.. Are you going to remove dataset list prod_*.txt file?

jshlee:
not straight away, but i think it's nicer to have all info in one place.

jhgoh:
Original idea of catListDatasets script was to list up xrood from the prod__.txt file, before filling up the spreadsheet. Theose prod__.txt files are directly used in the production, right?

jshlee:
for v745 we did production by reading the spread sheet and not the text files.

jhgoh:
OK. it's now clear. BTW, how about having another tab for Name:DataSetName:title:colour, etc? I think these information are unchanged in different campaigns.

PR #303, opened 2015-11-17
upgrade histogram drawing tool

tt8888tt:
by using histoHelper.py, made drawing easiler.
(also edited histoHelper.py to optimize size and offset)


jshlee:
hey dj, can u update this pr

PR #307, opened 2015-11-23
for v7-4-6 production

jshlee:
added both silver and golden weights
added flag in runData for silver and golden into pileweights
updated jet scale factors, JER and btagging
update electron eff area


jhgoh:
Why don't you just have another producer? The lumi selection bit is not "pileup weight".

jshlee:
i just put it as pileup weight for convenience, so the same variable can be used for both data and mc.

jhgoh:
Maybe this would be less confusing with "lumiWeight".

CatProducer/plugins/PileupWeightProducer.cc
```diff
@@ -106,9 +107,9 @@ PileupWeightProducer::PileupWeightProducer(const edm::ParameterSet& pset)
 void PileupWeightProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup)
 {
   std::auto_ptr<int> nTrueIntr(new int(-1));
-  std::auto_ptr<float> weight(new float(1.));
-  std::auto_ptr<float> weightUp(new float(1.));
-  std::auto_ptr<float> weightDn(new float(1.));
+  std::auto_ptr<float> weight(new float(0.));
+  std::auto_ptr<float> weightUp(new float(0.));
+  std::auto_ptr<float> weightDn(new float(0.));
```
jhgoh: Should be float(1.)

DataFormats/src/Jet.cc
```diff
@@ -69,85 +72,106 @@ float Jet::scaleFactorCSVv2(Jet::BTAGCSV_CUT cutType, int syst, JETFLAV flav) co
...
+  if (cutType == 0 && syst == 0 && flav == 2 && pt > 20 && pt < 1000 ) return ((1.07278+(0.000535714*pt))+(-1.14886e-06*(pt*pt)))+(7.0636e-10*(pt*(pt*pt)));
+  if (cutType == 0 && syst == -1 && flav == 2 && pt > 20 && pt < 1000 ) return ((1.01637+(0.000265653*pt))+(-4.22531e-07*(pt*pt)))+(2.23396e-10*(pt*(pt*pt)));
+  if (cutType == 0 && syst == +1 && flav == 2 && pt > 20 && pt < 1000 ) return ((1.12921+(0.000804962*pt))+(-1.87332e-06*(pt*pt)))+(1.18864e-09*(pt*(pt*pt)));
+  if (cutType == 1 && syst == 0 && flav == 2 && pt > 20 && pt < 1000 ) return 1.14022;
+  if (cutType == 1 && syst == -1 && flav == 2 && pt > 20 && pt < 1000 ) return 0.94022;
+  if (cutType == 1 && syst == +1 && flav == 2 && pt > 20 && pt < 1000 ) return 1.34022;
+  if (cutType == 2 && syst == 0 && flav == 2 && pt > 20 && pt < 1000 ) return 0.907317;
+  if (cutType == 2 && syst == -1 && flav == 2 && pt > 20 && pt < 1000 ) return 0.557317;
+  if (cutType == 2 && syst == +1 && flav == 2 && pt > 20 && pt < 1000 ) return 1.257317;
```
jhgoh: Should be pt >= 20 etc.
Do you think that this form is better than the previous one?

jshlee: the previous one is better, but this was much easier to implement from their excel sheet.

PR #309, opened 2015-11-26
h2muDraw updates

pseudotop:
added -f option for renaming in h2muDraw.py.
and "run_h2muDraw.py" file for running batch job of h2muDraw.py.
I made "set_config.py" for a variety of variables and cuts.
"merge2Canvas" is made for just Jet Categoriztion.


jhgoh:
This is an analysis tools update mostly on plotting. Merging this is safe and the change is to be monitored by the analyzer.

PR #313, opened 2015-12-02
add pseudoTop info to compare with recoTop

tt8888tt:
pseudo Top and TTbar infos has been added to tree.
changed some values to sync condition for muon & bjet (commented out).
for topdaw.py, just minor details changed.


CatAnalyzer/plugins/TtbarDiLeptonAnalyzer.cc
```diff
@@ -555,9 +612,9 @@ void TtbarDiLeptonAnalyzer::selectMuons(const cat::MuonCollection& muons, Partic
   for (auto& mu : muons) {
     if (mu.pt() < 20.) continue;
     if (std::abs(mu.eta()) > 2.4) continue;
-    //if (!mu.isMediumMuon()) continue;
     if (!mu.isTightMuon()) continue;
     if (mu.relIso(0.4) > 0.12) continue;
+    //if (mu.relIso(0.4) > 0.15) continue;//forsync
```
jshlee: check if this is correct.
0.12 might be an old value

Issue #311, opened 2015-11-27
To check that the secondary vertex collection can be embeded.


geonmo:
### Motivation
- For J/psi analysis,  I want to embed the secondary vertex collection to official CATTools.
  However, in most case, the secondary vertiecs(below SV) have too big size to put the ntuple.
  So, we need to check their size, precut with least loss and input collection to reconstruct to SV.
### CheckPoint ( Higher is more important. )
1. Size           : small size is mandatory for official production.
2. Efficiency  : can be produced using miniAOD.
3. Availability : can be produced using miniAOD.
4. Time          : too many time cause to fail grid job.
### Methods
1. <strike>Conventional Lepton+Lepton reconstruction</strike> : Obsoleted by 8TeV Analysis due to too small statistics
2. Lepton + general tracks( Now, patMuons + generalTracks )
   - GeneralTracks can be available on AOD samples.
   - Already porting to CMSSW 7XY.
3. MiniAOD's SlimmedSecondaryVertices
   - Available with miniAOD.
   - It did not produced for j/psi. So, I need to check about efficiency of J/psi
4. packedPFCandidates + packedPFCandidates
   - Available with miniAOD.
   - It also need to check of J/psi efficiency
### Time Table( Higher is first )
1. As soon as possible to porting (3), (4) reco methods as CATTools producer.
2. Compare the size between 2,3 and4
3. Compare the efficiency between 2,3 and4

geonmo:
#### Quick & Dirty results
-  Size of method(2)

``` bash
recoGenParticles_prunedGenParticles__CAT. 3950.91 781.356
catSecVertexV2s_catSecVertexsV2__CAT. 4375.59 1384.51
catSecVertexs_catSecVertexs__CAT. 131.93 25.799
```
- Maybe, catSecVertex**V2** **can not be used** for cat ntuple without strong precut. ( x4 catJets ) :boom:

jhgoh:
what is catSecVertexsV2?

jshlee:
i think catSecVertexsV2 is the updated one

jshlee:
@geonmo use the cern cuts for now
and also add in D\* candidates

geonmo:
@jhgoh catSecVertexV2 means 2nd methods.(Lepton + generalTracks)
I will also make V3 and V4 for method 3 and 4.

geonmo:
Of course, it will not push to official. These are made to compare J/psi candidates for each modules.

jhgoh:
[...]

PR #317, opened 2015-12-03
Modify CATSecVertex from Lepton+Lepton to pfCandidate+pfCandidates.

geonmo:
### Introduction

On issue #311, A new method which make a secvtx candidates from pfCandidates is better efficiency than conventional method. So, I will replace old to new method for official catSecVtx producer.
### Feature
- Replace CATSecVtx to pfCandidate method.
- Add legacy plugins for CatAnalyzer's catJpsiDraw directory.
- Add some scripts to draw and compare producer's performance.


jshlee:
are CatAnalyzer/test/catJpsiDraw/plugins/CATSecVertexXXProducer.cc required?

geonmo:
@jshlee No, it did not required. It is just for code review.

jhgoh:
Why vertex fit is there for V3? I thought just copying secondary vertex collection is enough... was it assuming pion legs??

jshlee:
i think u can delete all the different versions and have it all in one plugin with different options

geonmo:
@jhgoh Actually, no reason for that. It is just due to cloning from original code.
@jshlee Yes, you are right. I will put them all to one plugin next week.

jshlee:
what is the status of D*?

PR #320, opened 2015-12-10
Histogram cut flow analyzer

jhgoh:
This PR modifies TTbarDileptonEventSelector (with new name TTLLEventSelector) to make most of control plots with very fine binning (1GeV for pt, mass distributions) for each pre-defined cut flows.
The PU weight and gen weight are applied by default for the MC samples (set "isMC" flag in the cfg file). Other scale factors are not implemented in this version.

The module produces Ptrs to original objects of selected leptons and jets thus one can append other analyzer/producers such as kinematic reconstruction module (TTLLKinSolutionProducer), the GenericNtupleMaker, etc.

The "channel" information is also available, determined using the leading two lepton's flavour.


jshlee:
ready to merge?

jhgoh:
Not in hurry, I'm going to merge with some additions.

jhgoh:
This PR includes first part of histogram based analysis workflow. Histograms are produced without problem and we can proceed to the next step.

PR #330, opened 2015-12-22
Ttbb run2

monttj:
This update is to use genTop for ttbb analysis.
It is OK now for dilepton decay mode.
But we need more updates for lepton plus jets as we need to remove jets from W in lepton plus jets mode.


CatProducer/prod/PAT2CAT_cfg.py
```diff
@@ -8,7 +8,7 @@
 options.register('runOnMC', True, VarParsing.multiplicity.singleton, VarParsing.varType.bool, "runOnMC: 1  default")
 options.register('useMiniAOD', True, VarParsing.multiplicity.singleton, VarParsing.varType.bool, "useMiniAOD: 1  default")
 options.register('globalTag', '', VarParsing.multiplicity.singleton, VarParsing.varType.string, "globalTag: 1  default")
-options.register('runGenTop', False, VarParsing.multiplicity.singleton, VarParsing.varType.bool, "runGenTop: 1  default")
+options.register('runGenTop', True, VarParsing.multiplicity.singleton, VarParsing.varType.bool, "runGenTop: 1  default")
```
jhgoh, line 11: I think default should be False for safety.
This option can be turned on with cmsRun arguments.

PR #333, opened 2015-12-23
added sf for leptons

jshlee:
removed flavour input for btagging sf
updated most systematics for ttbar analysis


jhgoh:
This PR is for the scale factors to be used in the analysis. There are dataformat changes but they are just for const functions to return SF's, no change in dataformat content.

PR #339, opened 2015-12-29
Cleaner dataset txt file management

jhgoh:
dataset_*.txt files are saved under a subdirectory not to make data directory 2b messy.


jshlee:
im ok with this

PR #344, opened 2015-12-30
Add SF evaluator to CommonTools

jhgoh:
Add scale factor evaluator.


jhgoh:
For note to @tt8888tt , @kuyoun : This PR also changes the TtbarDileptonAnalyzer to split muon SF variation and electron SF variation.

PR #349, opened 2016-01-08
change analyzer code (branch names)

tt8888tt:
combined muon SF up & down and electron SF up & down to lepton SF up & down.


jhgoh:
Can we simply assume 100% positive correlation bet'n mu scales and electron scales?

tt8888tt:
You mean like... up SF for electron is saved as 'lepton_weght' in branch 'mueff_u' too...?

jhgoh:
Previous code saves weights from mu and electrons separately.

jshlee:
for ee and mumu channel, it shouldnt matter, but for emu channel, this wouldn't work.
i dont think there will be 100% positive correlation, so e and mu should be done separately.

jhgoh:
Codes are up to date to reflect those comments.

PR #352, opened 2016-01-11
Completing TTLL chain

jhgoh:
(hopefully) this batch of PR is to finalize the histogram based workflow.

The RooStats are introduced, systematic errors are considered to extract numbers.
Analysis specific plots are going to be introduced.


jhgoh:
Now the RooStats/HistFactory gives some -profile LL curve with histogram based syst. errs. NLL's are not smooth at some points - to be checked. Fit was done using the leading jet's b discrinimator distribution JUST FOR TESTING.

Updated version with TTBB ratio can be practically useful..

meas_combined_xsec_profilelr eps
meas_ee_xsec_profilelr eps
meas_em_xsec_profilelr eps
meas_mm_xsec_profilelr eps

jhgoh:
This PR is not 100% complete, but basic workflow works anyway. Since we are moving to 76x branch so it's time to finalize this PR and move to the next branch.

PR #354, opened 2016-01-12
Add photonto cattuples

jalmond:


jhgoh:
Codes in good shape. Merging this PR.

PR #356, opened 2016-01-14
Top pt reweight

jhgoh:
Top pt reweighting factors are produced by the PartonTopProducer.


jhgoh:
Adding production flag since this module runs during the prod.

PR #357, opened 2016-01-14
a bit change of histohelper.py

tt8888tt:
make const k be calculated separately in step 1, once only.


jhgoh:
Merge?

jshlee:
hey @tt8888tt 
why is kMM and kEE calculated separately?

tt8888tt:
cause I've learned that I have to calculate kMM, kEE at Step1(before MET cut) and use that in all other steps.
or is it wrong?

jhgoh:
Some discussion done yesterday, he was just asking about technical stuffs something like why dont you calculate k factors on the fly.

tt8888tt:
ah then, it's because I just simmply don't know how to, technically.
can't figure out better way...

jhgoh:
Merging this PR. no harm to be merge.

PR #373, opened 2016-01-22
Btag weight interface

jhgoh:
Adding an interface to use CSV weight.
This interface is taking the CSV weights from the .csv file using the common interface given by the B tag POG.
The interface itself is designed for the other methods too.


jhgoh:
Now the code is tested with @YoungKwonJo, runs no crash.

PR #394, opened 2016-01-28
catGetDatasetInfo update

jhgoh:
catGetDatasetInfo script is updated
- Default prod version v7-6-1
- Retrieve sample list first and try to run xrdls later - works at CERN
- Human readable JSON file


jshlee:
+1

PR #395, opened 2016-01-28
updated ttdilep anlaysis

jshlee:
changed back to cutBasedElectronID
updated working point for btagging in 76x

need to wait fot new sf for electrons and btagging


jhgoh:
Adding production tag since there is some change in input txt files

PR #412, opened 2016-02-07
Include central weight at the first of all weight variations

jhgoh:
Keep the 0th element of weight group.
This can be a duplication of information, but it gives uniform structure for all variations.

Therefore, genWeight:genWeight = genWeight:scaleWeights[0] = genWeight:pdfWeights[0] = genWeight:otherWeights[0](if exists)


jshlee:
can this be merged?

jhgoh:
Merging this.

PR #413, opened 2016-02-07
76X Working points

jhgoh:
Use common header file to define various working points.


DataFormats/interface/Jet.h
```diff
@@ -7,6 +7,8 @@
 #include "DataFormats/PatCandidates/interface/Jet.h"
 #include "DataFormats/PatCandidates/interface/JetCorrFactors.h"
 
+#include "CATTools/CommonTools/interface/WorkingPointDefs.h"
+
```
jshlee: why dont we just add in the working points to this header?

jhgoh: yeah it would be better.

jshlee:
+1

PR #416, opened 2016-02-08
KinSolver angle smearing from DESY code

jhgoh:
Took codes from DESY group on the angle smearing.
I was not adding this just because the angle smearing approximately holds with the previous implementation, rotation in theta-phi space.
We are trying to be as close as possible to the reference.

To be merged AFTER TESTING.


jhgoh:
Commits after 320efa9 reflects PR #419 thus electron SF takes super cluster eta. It should not affect top kinreco itself.

jhgoh:
Merging this.

PR #419, opened 2016-02-08
Keep inheritence information

jhgoh:
SelectMuons and SelectElecs functions saved selected leptons in LeptonCollection for simplicity but this looses inheritance information thus calling dynamic_cast failed.

In this fix, those functions stores selected object in the collections with same type and their pointers are uses in pT ordering.


jhgoh:
@YoungKwonJo tested with small sample and there was no difference with this fix.

PR #423, opened 2016-02-09
add a folder for ttbb dilepton ntuple production

YoungKwonJo:


jhgoh:
Can you give another name for the create-batch2 script?

PR #441, opened 2016-02-18
For csv helper

YoungKwonJo:
I found a strange bug in "CatAnalyzer/src/BTagWeightEvaluator.cc".
It was compiled. however for based on root-files, it is getting wrong event weight value.

So I would like to use the old-one(my used) as CSVHelper.
CSVHelper gives a correct value. And, the following link is original code. I would like to keep the coding style. because it is not easy to debugging.
https://github.com/cms-ttH/MiniAOD/blob/master/MiniAODHelper/interface/CSVHelper.h


YoungKwonJo:
And,now, the SF is not same between csv-file and root-files. so we need to see more.
btagsf_rootvscsvfile

jhgoh:
Difference between csv file and root file is documented in the twiki page, but I don't know how it differs.

jhgoh:
BTW, can you summarize what is the issue of wrong weighting factors?

jhgoh:
Actually, the CSV/root file part of the BTagWeightEvaluator is just copy and paste of CSVHelper. What is the problem?

YoungKwonJo:
no, CSVHelper is not using enum. perhaps, we have to see more hardly for BTagWeightEvaluator. 
And, the event weight is normally almost 1. however BTagWeightEvaluator with option root-file gave big number. so we need to debug.

YoungKwonJo:
I guess we need to discuss by offline. I guess the effect of the difference is not big for our result and shape. however they have a little different.
I have checked the result(analysis) by using both as root-file and csv-file.

PR #443, opened 2016-02-20
New lepton sf for 76X and 74X recalculation

jhgoh:
Official muon scale factors are updated.
The old 74X scale factors are renamed to muonSFTight74X.

Same for the electrons, they are updated from Feb 19
https://indico.cern.ch/event/491528/contribution/2/attachments/1231399/1805319/CutBasedId_EGM_19Feb.pdf
with the new txt format to follow the EGammaPOG common format. The uncertainty propagation is done again based on the new format.
Existing 74X scale factors are re-calculated based on new format as well.

Note for the analyzers: electron SF for 76X is given by ABSOLUTE ETA. Analysis modules have to be modified to read from abseta_bins and the analysis modules have to be modified correspondingly.


jhgoh:
analyzer modules are updated to use abseta for electrons.

jshlee:
merge?

PR #449, opened 2016-02-25
For csv helper

YoungKwonJo:
added CSVHelper.h, cc


jshlee:
what is the status of this???
you need to fix conflicts too

jhgoh:
I and Youngkwon had discussion on this issue. This can be merged after resolving conflict, maybe in the afternoon.
This is on the analysis level, thus it can be merged any time.

Issue #434, opened 2016-02-11
prod 7-6-3


jshlee:
- [x]  once the new JEC is production i will start a new prod, fall15 based jec should be out soon
- [ ]  MET - met corrections are not working. can turn off till fix comes out. this just means no noHF met and no corrections. but since it is using same jec miniaod met can be used out of box. still waiting for fix from met group
- [x]  update data samples -- waiting for more Fall15 samples to be production. currently 17% is missing but not urgent
- [ ]  photon energy smearing - electron energy smearing is already applied, waiting for similar one for photons

jshlee:
done!

PR #450, opened 2016-02-25
Test sample for 763 prod

jhgoh:
catTuples for testing are copied to eos.
Updated analyzer cfg files to point new location.


jhgoh:
This is one of the final step of #434

PR #452, opened 2016-02-26
drawing tool update

tt8888tt:
optimized a bit drawing tools.
added combined channel drawing script and command print out script.


jshlee:
analysis level only
no compile needed
merged

PR #454, opened 2016-02-28
ttll analysis

jshlee:
reduced ttrees based on weights


jshlee:
merging because this is analysis only

PR #456, opened 2016-02-29
genWeight index bugfix

jhgoh:
There was problem to treat GenWeight indices. Indices can be shifted by one or more.


jhgoh:
To be merged after quick tests

jhgoh:
ScaleWeight[0] before fix
canvas_1

ScaleWeight[0] after fix

PR #458, opened 2016-02-29
Update unit test input

jhgoh:
input root files for the unit test is updated to 762 relval.


jhgoh:
This PR is just for the local tests, safe to be merged.

PR #459, opened 2016-03-01
Gen jet hadron matching information with decay history

jhgoh:
Per-genJet hadron matching information is stored in the output.
Also, the list of parton's pdg ids are stored to do the parton-genJet matching, just keeping pdgIds of interested partons (top, W,Z H).

The output size is expected to be small.

```
Type                                  Module                   Label             Process   
-------------------------------------------------------------------------------------------
vector<int>                           "genJetHadronFlavour"    "flavour"         "CAT"     
vector<vector<int> >                  "genJetHadronFlavour"    "ancestorIds"     "CAT"     
```

```
Branch Name | Average Uncompressed Size (Bytes/Event) | Average Compressed Size (Bytes/Event) 
intss_genJetHadronFlavour_ancestorIds_CAT. 122.29 36.83
ints_genJetHadronFlavour_flavour_CAT. 73.63 30.53
```

@YoungKwonJo please check this is what you wanted.

### DIFF
--- a/CatProducer/plugins/CATGenJetProducer.cc
+++ b/CatProducer/plugins/CATGenJetProducer.cc
@@ -3,20 +3,11 @@
 #include "FWCore/Framework/interface/Event.h"
 #include "FWCore/ParameterSet/interface/ParameterSet.h"
 
-#include "DataFormats/Common/interface/Association.h"
-#include "DataFormats/Common/interface/RefToPtr.h"
-
-#include "DataFormats/JetReco/interface/GenJetCollection.h"
-#include "DataFormats/JetReco/interface/GenJet.h"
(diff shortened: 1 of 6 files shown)


jhgoh:
It is worth to have number of matched hadrons. Also the last commit adds parton pdg ids from matched HF hadrons.

```
intss_genJetHadronFlavour_ancestors_CAT. 122.01 36.55
intss_genJetHadronFlavour_hadronAncestors_CAT. 88.61 33.79
ints_genJetHadronFlavour_nCHadron_CAT. 73.77 30.55
ints_genJetHadronFlavour_nBHadron_CAT. 73.77 29.83
```

PR #461, opened 2016-03-02
BTagWeightEvaluator code reduction

jhgoh:
The CSVHelper is imported to CATTools by PR #449, thus the code fragments in the BTagWeightEvaluator becomes obsolete thus removed.
The CSVHelper code will be maintained by @YoungKwonJo.

This class still can be used as a generic interface to apply various btagging SF applications.


jhgoh:
I'll add one more commit to make a wrapper to the CSVHelper.

PR #465, opened 2016-03-05
Smearing only if exact solution is missing

jhgoh:
Take the solution without smearing if it exact solution exists - this gives better resolution.
- Top quark (pt_kin - pt_truth)/(pt_truth)
  htdpt
- Top quark (eta_kin - eta_truth)
  htdeta
- Top quark (phi_kin - phi_truth)
  htdphi

This is to be merged after confirmation.


jhgoh:
I'm closing this for now as #466 covers this issue. Maybe CATTools-compatible implementation can be done later.

PR #469, opened 2016-03-09
dos2unix

jhgoh:
Nothing but just remove dos-based new line character.


jshlee:
but this is what makes desy kin fit work!!!

PR #470, opened 2016-03-09
Use XYZTLorentzVector for compatibility

jhgoh:
DESY kinsolver extensively uses PtEtaPhiM interpretation to return the results, but math::XYZTLorentzVector is the standard format to store lorentzVector.

Migrating to use the PxPyPzE reduces un-necessary conversions between them.
There should no physics change up to numerical precision (in fact, this migration reduces numerical error propagation).


jhgoh:
This PR is included in #475

PR #471, opened 2016-03-09
Random number usage in CC analysis

jhgoh:
Bugfix to pick random number.


jhgoh:
@hyunyong , please double check the codes.

PR #475, opened 2016-03-10
Top kin solver bugfix

jhgoh:
The common top kin solver had bugs
- pi was set to wrong value (most stupid bug during the kin solver)
- lepton index is flipped (this involves most complicated set of equations thus it took quite long time to find the problem)

In the DESY's original implementation, top quark mass is enforced to the input top quark mass after smearing. The solver is also modified.


geonmo:
현재까지 별 문제가 없는 것 같으면 merge 하겠습니다.

jhgoh:
efficiency 에 차이가 있다고 해서 약간 더 디버깅이 필요해요.

PR #463, opened 2016-03-04
GenWeight dataformat

jhgoh:
This PR adds new dataformat for generator level weights - packs up genWeight stuffs into member variables.

Main features:
- Scale weights are split into "scaleUp" and "scaleDown" and the unphysical up+down combinations are dropped, thus users can easily apply scale weights in their analysis codes.
- Parameters for each weight items are stored in the run, GenWeightInfo class to retrieve the information.

```
edmDumpEventContent --run catTuple.root 
Type                   Module        Label     Process   
---------------------------------------------------------
cat::GenWeightInfo     "genWeight"   ""        "CAT"     
```

```
edmDumpEventContent  catTuple.root 
Type                                  Module                   Label             Process   
-------------------------------------------------------------------------------------------
cat::GenWeights                       "genWeight"              ""                "CAT"     
```

Size is somehow large, but I think we have much more gain in convenience.
catGenWeights_genWeight__CAT. 1275.31 905.21

PR #485, opened 2016-03-17
Better scale weight

jhgoh:
Reopen #463 To be merged just before the prod starts.


jhgoh:
(re) merging this PR before the production starts.
Analyzers can use the tag v7-6-5 if one needs setup without GenWeights dataformat.

PR #487, opened 2016-03-17
KinSolver bugfix + Priority to solution with b tagged jet

jhgoh:
Give priority to the solution with more b tagged jets.
In addition to this, at least one b tagged jet is required to be included in the kinematic solution.

To enable these selections, add a struct to save KinSolver result. This makes analyzers to manager kinsolver results in their analysis module.

### DIFF
--- a/CatAnalyzer/interface/KinematicSolvers.h
+++ b/CatAnalyzer/interface/KinematicSolvers.h
@@ -67,6 +67,11 @@ class KinematicSolver
 public:
   KinematicSolver(const edm::ParameterSet& pset) {}
   virtual ~KinematicSolver() {}
+  void solve(const LV met, const LV l1, const LV l2, const LV j1, const LV j2)
+  {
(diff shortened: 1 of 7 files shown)


jhgoh:
This PR is not complete yet. to be merged after bugfix in this PR.

jhgoh:
Made some bugfix, synchronized numerical precision, applied b tagged jet preference.
top quark pt difference, rapidity difference and ttbar pt difference to the parton level object is given below, red is the one from this implementation and the blue line is from the DESY's original code.

topptmax
topptmin
topymax
topymin
ttpt

jhgoh:
There's one thing I actually don't understand at all - sometimes the solver fails to reproduce the original result when lepton and anti-lepton's ordering is flipped. 
If things are ideal, lepton1 and lepton2 should be symmetric since all pairs are tested and the pair with maximum weight is selected. I'm assuming the difference comes from numerical error while finding roots. Those effects must be wiped out with random smeareaings.

PR #505, opened 2016-03-28
correct orderring values in analyzer

tt8888tt:
make every values in charge order
and add nutrino 4vec on ttree
and update plot drawing tools


jshlee:
merge?

PR #506, opened 2016-03-28
changed drawing tool as value names changed on tree

tt8888tt:


geonmo:
This commit only affect to private code.

PR #514, opened 2016-03-30
Add processedLumis

geonmo:


jshlee:
is this the best way for us to organise processed lumi?

jhgoh:
These processedLumi file is to be used to (re) calculate lumi for each datasets.
They can be placed in the xrootd directory where root files are placed, maybe with the file list in txt format.
Otherwise, we can create new repo to store those prod information and put the link to the repo as a submodule.

geonmo:
Closed this PR.

jshlee:
i prefer to have the processedLumi file in the xrootd directory where root files are placed
i would like a more automated solution to this

PR #515, opened 2016-03-31
Add submodule for dataset management

jhgoh:
**Motivation**:
Information retrieved from gDoc can be changed without notification thus there is some risk of miscommunication.
Sometimes gDoc can be screwed up due to mistake. Uses can retrieve date during editing, etc.

**Suggested change from this PR:**
For the end-users - provide centrally produced json file and file list txt files. Put them in dedicated git repo for the production archiving.
https://github.com/vallot/CATTools-dataset
The sample information will be version-controlled.

**Longer term plan:**
Produce some web interface to give summary table for the users.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "CommonTools/scripts"]
 	path = CommonTools/scripts
 	url = https://github.com/cms-kr/hep-tools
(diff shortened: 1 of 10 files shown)


geonmo:
I think that it is best for now.

Issue #516, opened 2016-04-04
GenWeight problem in MadGraph samples


jhgoh:
GenWeight is not filled in correct way for the MadGraph samples.
This is because the names of weight groups are not uniform to the powheg and aMC@NLO.

All scale variations, PDF variations went to the otherweights in these samples.

jhgoh:
Moving back to this issue.
You can have a look into detailed by accessing catGenWeightInfo_genWeight__CAT.obj.print(); in the Runs tree.

I'm adding temporary solution to this problem, just to add additional string matching criteria dedicated to this generator.
But there should be general solution to sort out them.

```
root [3] ------------------------------------------------
Central scale variation envelope
  ->  mur=1 muf=1  access using otherWeights[0]
  ->  mur=1 muf=2  access using otherWeights[1]
  ->  mur=1 muf=0.5  access using otherWeights[2]
  ->  mur=2 muf=1  access using otherWeights[3]
  ->  mur=2 muf=2  access using otherWeights[4]
  ->  mur=2 muf=0.5  access using otherWeights[5]
  ->  mur=0.5 muf=1  access using otherWeights[6]
  ->  mur=0.5 muf=2  access using otherWeights[7]
  ->  mur=0.5 muf=0.5  access using otherWeights[8]
NNPDF30_lo_as_0130.LHgrid hessian
  -> Member 0 access using otherWeights[9]
```

jhgoh:
Closing this issue.

PR #519, opened 2016-04-23
Quick fix for the correct scale variation weights

jhgoh:
GenWeight for the madgraph samples were not correctly ordered in 764 samples. #516 
This is because each generators have different lhe header tags and attributes to store the weight information.

Users can avoid the problem simply forcing the weights to be 1 for the madgraph samples or get the weights from the otherWeights vector.


geonmo:
+1

geonmo:
I think that it can merged now.

PR #524, opened 2016-05-06
Simplified GenWeight dataformat

jhgoh:
cat::GenWeightInfo and cat::GenWeight are stored in the catTuple but the dataformat is little bit simplified not to split weight vector during the production step.

Categorizing weights is done on the fly by an additional module during ntuple production, end users can read flat formats.

screen shot 2016-05-07 at 3 25 59 am

This is safe from possible LHE header changes, one can modify the flat weight producer and rerun the ntuplers. catTuple can remain as is.


jhgoh:
Also, scale up/down weights are split by their parameter in LHE header, simply looking for strings with "5" or "2". If both "5" and "2" appeared from an weight, it is considered as an unphysical combination. No parting is done, it is just a simple string matching.

jhgoh:
@geonmo , please include this PR if you did not start new production.

jhgoh:
Migrated analyzer codes to run on updated dataformat.

PR #529, opened 2016-05-17
Add Jpsi candidates to D* producer.

geonmo:


CatProducer/plugins/CATDStarProducer.cc
```diff
@@ -111,6 +119,95 @@ cat::CATDStarProducer::produce(edm::Event & iEvent, const edm::EventSetup & iSet
...
+
+    for ( unsigned int lep1_idx = 0 ; lep1_idx< dau_size-1 ; lep1_idx++) {
+      for ( unsigned int lep2_idx = lep1_idx+1 ; lep2_idx< dau_size ; lep2_idx++) {
+        bool flag_jpsi = true;
+        Shared_PCP lep1Cand = jetDaughters[lep1_idx];
+        Shared_PCP lep2Cand = jetDaughters[lep2_idx];
+
+        int pdgMul = lep1Cand->pdgId() * lep2Cand->pdgId();
+        if ( pdgMul != -121 && pdgMul != -169 ) flag_jpsi = false;
+        else if ( !flag_jpsi && abs( pdgMul ) !=2321 && abs( pdgMul) != 2743 ) continue;
```
jhgoh: flag_jpsi is always true in this else-if statement.
So the statement `!flag_jpsi` will give false and this line will not be executed.

geonmo:
@jhgoh Hello, I fix the bug. I took a mistake.

Issue #535, opened 2016-05-21
2016 run data production


jshlee:
need to update for catDefinitions_cfi.py
- [x]  JECUncertaintyFile
- [x] lumiJSON
- [x] pileupMCmap - need to add in 2016_25ns_SpringMC_PUScenarioV1_PoissonOOTPU
- [x] JECUncertaintyFile

jhgoh:
JSON file was announced here:
https://hypernews.cern.ch/HyperNews/CMS/get/physics-validation/2648.html

Luminosity calculation using the unofficial normtag now returns 211.031 pb-1.

```
$ runLumiCalc lumi normtag_DATACERT.json /afs/cern.ch/cms/CAF/CMSCOMM/COMM_DQM/certification/Collisions16/13TeV/Cert_271036-273450_13TeV_PromptReco_Collisions16_JSON_NoL1T.txt 
Running brilcalc
brilcalc lumi --normtag /afs/cern.ch/user/l/lumipro/public/normtag_file/normtag_DATACERT.json -i /afs/cern.ch/cms/CAF/CMSCOMM/COMM_DQM/certification/Collisions16/13TeV/Cert_271036-273450_13TeV_PromptReco_Collisions16_JSON_NoL1T.txt
#Data tag : online , Norm tag: composite
+-------------+-------------------+------+------+----------------+---------------+
| run:fill    | time              | nls  | ncms | delivered(/ub) | recorded(/ub) |
+-------------+-------------------+------+------+----------------+---------------+
| 273158:4915 | 05/11/16 20:42:35 | 1279 | 1279 | 53933772.772   | 50633622.283  |
| 273291:4919 | 05/12/16 20:23:51 | 77   | 77   | 3273256.636    | 2757452.659   |
| 273299:4919 | 05/12/16 21:55:32 | 43   | 43   | 1867156.107    | 1747801.907   |
| 273301:4919 | 05/12/16 22:14:12 | 602  | 602  | 24646788.835   | 23465019.864  |
| 273302:4919 | 05/13/16 02:13:38 | 459  | 459  | 17074774.630   | 16380269.045  |
| 273402:4924 | 05/14/16 01:49:19 | 193  | 193  | 7006325.061    | 6628630.533   |
| 273403:4924 | 05/14/16 03:05:27 | 53   | 53   | 1932155.788    | 1834285.375   |
| 273404:4924 | 05/14/16 03:31:50 | 18   | 18   | 651633.249     | 615370.075    |
[...]

PR #540, opened 2016-05-26
Spring16 Pileup and MC JEC

jhgoh:
JEC db file for the Spring16 production. #535
https://twiki.cern.ch/twiki/bin/view/CMS/JECDataMC
This is already applied in MiniAODv2.

MiniAODv1 is based on 76X.

JEC for the Data is not appeared in the twiki yet (May26)
JEC Uncertainty can be retrieved from the same twiki page, but not the one with different uncertainty sources.


jhgoh:
This PR may crash until having appropricate JECUncertainty files.

jhgoh:
New PU weights will be added with new recommendation of minBias cross section.
https://hypernews.cern.ch/HyperNews/CMS/get/luminosity/594.html

PR #541, opened 2016-05-27
New package: validation

jhgoh:
Requesting to add new package for the CATTuple validation.
As a first step, set of generator information are added.

The first batch of jobs are submitted to test the 765 production.
Average genWeight will be obtained.


jhgoh:
First look on the plot(s)

screenshot

PR #550, opened 2016-06-01
use the reference trigger selection

jhgoh:
One of the trigger path was missing in the cfg file.
This is necessary to use the reference trigger scale factors.


jhgoh:
@tt8888tt @kuyoun @jshlee note that the trigger was missing.
Please consider to redo the ntupling job with new cfg.

PR #553, opened 2016-06-02
update GenWeightsProducer

YoungKwonJo:
for safe in using Real DATA.


YoungKwonJo:
Still, it is not fixed.

PR #554, opened 2016-06-02
Generic way to split genWeights

jhgoh:
Now the scale weights are split based on the actual muR and muF parameters in generic way.
Some basic level of string parsing was done to determine the scale parameters.


jhgoh:
Tested by Youngkwon. Merging this PR.

PR #556, opened 2016-06-09
use default miniAOD jet pt and fix pileup using 2015 (needs to be fixed)

monttj:
1. pileup weight needs to be updated.
   currently it is using 2015 weight just to make the code work
2. use default MINIAOD jet pt unless updating jet pt is really needed.

### DIFF
--- a/CatProducer/prod/PAT2CAT_cfg.py
+++ b/CatProducer/prod/PAT2CAT_cfg.py
@@ -65,6 +65,22 @@ miniAOD_customizeOutput(process.catOut)
     
 process.catOutpath = cms.EndPath(process.catOut)    
 process.schedule.append(process.catOutpath)
+
+
+#### pileup weight
+import CATTools.CatProducer.catDefinitions_cfi as cat
+lumiJSON = "Cert_246908-260627_13TeV_PromptReco_Collisions15_25ns_JSON"
+if runOnMC:
+        from CATTools.CatProducer.pileupWeight_cff import pileupWeightMap
+        process.pileupWeight.pileupMC = pileupWeightMap[cat.pileupMCmap]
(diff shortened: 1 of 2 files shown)


CatProducer/python/catTools_cff.py
```diff
@@ -2,16 +2,6 @@
...
-        process.pileupWeight.pileupMC = pileupWeightMap[cat.pileupMCmap]
-        process.pileupWeight.pileupRD = pileupWeightMap["%s"%cat.lumiJSON]
-        process.pileupWeight.pileupUp = pileupWeightMap["%s_Up"%cat.lumiJSON]
-        process.pileupWeight.pileupDn = pileupWeightMap["%s_Dn"%cat.lumiJSON]
-    else:
-        from FWCore.PythonUtilities.LumiList import LumiList
-        process.lumiMask = cms.EDProducer("LumiMaskProducer",
-            LumiSections = LumiList('../data/LumiMask/%s.txt'%cat.lumiJSON).getVLuminosityBlockRange())
     
     useJECfile = True
```
jshlee, line 6: to turn off jec correction, you can set false here

CatProducer/prod/PAT2CAT_cfg.py
```diff
@@ -65,6 +65,22 @@
     
 process.catOutpath = cms.EndPath(process.catOut)    
 process.schedule.append(process.catOutpath)
+
+
+#### pileup weight
+import CATTools.CatProducer.catDefinitions_cfi as cat
+lumiJSON = "Cert_246908-260627_13TeV_PromptReco_Collisions15_25ns_JSON"
```
jshlee, line 72: i prefer to have the lumiJSON defined in catDefinitions_cfi, so all production info is in one place

jhgoh:
Closing this PR since there are other PRs that cover this.

Issue #577, opened 2016-06-24
Prod v8-0-0


jhgoh:
Being prepared for the 2016 data processing.

jhgoh:
https://docs.google.com/spreadsheets/d/1rWM3AlFKO8IJVaeoQkWZYWwSvicQ1QCXYSzH74QyZqE/edit#gid=1203220026

jhgoh:
Now all necessary recipes are implemented. Staring the production.

jhgoh:
Good to mention the vallot/cmssw PR https://github.com/vallot/cmssw/pull/15

jhgoh:
The latest JSON to cover all Run2016B is arrived and we also have Run2016C.
New JEC V6 is also available as well as the uncertaintySources txt file

The prod v800 is going to be resubmitted to include all changes.

jhgoh:
runtests OK, proceeding to the job submission.

jhgoh:
Job monitoring https://dashb-cms-job-task.cern.ch/dashboard/templates/task-analysis/#user=Junghwan+Goh&refresh=0&table=Mains&p=1&records=25&activemenu=2&pattern=160705_*jhgoh_crab_v8-0-0*&task=&from=2016-07-05&till=2016-07-06&timerange=lastMonth

jhgoh:
Job status check
- [x] crab_v8-0-0_DoubleMuon_Run2016C-PromptReco-v2
- [x] crab_v8-0-0_SinglePhoton_Run2016C-PromptReco-v2
- [x] crab_v8-0-0_ST_s-channel_4f_leptonDecays_13TeV-amcatnlo-pythia8_TuneCUETP8M1
- [x] crab_v8-0-0_ST_t-channel_antitop_4f_leptonDecays_13TeV-powheg-pythia8_TuneCUETP8M1
- [x] crab_v8-0-0_ST_t-channel_top_4f_leptonDecays_13TeV-powheg-pythia8_TuneCUETP8M1
- [x] crab_v8-0-0_ST_tW_antitop_5f_inclusiveDecays_13TeV-powheg-pythia8_TuneCUETP8M1
- [x] crab_v8-0-0_ST_tW_top_5f_inclusiveDecays_13TeV-powheg-pythia8_TuneCUETP8M1
- [x] crab_v8-0-0_ST_tW_top_5f_NoFullyHadronicDecays_13TeV-powheg_TuneCUETP8M1
[...]

PR #586, opened 2016-06-27
Forward port of GenTop fix

jhgoh:
Forward for of #585


jhgoh:
The ttbar samples for 800 prod campaign #577 have to be restarted.

PR #587, opened 2016-06-27
fixed bug in gentop

YoungKwonJo:
##


jhgoh:
Was it pass scram b runtests?

YoungKwonJo:
Yes, I tested it in the foloder **_CatProducer/test_** as like following.

> scram build runtests

PR #588, opened 2016-06-27
GenTop bugfix, crab submit scripts for 80X

jhgoh:
Forward port of #587 
Update CommonTools scripts
Minor fix on crabSubmit script


jhgoh:
To be applied #577 except the genTop stuffs yet.

PR #589, opened 2016-06-28
fix bug

monttj:
obvious bug and it is fixed now


jhgoh:
What was difference to #587? problems in git?

monttj:
It is the same. Then how come it is possible that it is not fixed yet?

PR #590, opened 2016-07-05
Updated recipe for the first 80x production

jhgoh:
There are massive update to process 80X data as of July 5.
This PR apply most recent recipes to restart 80X production. Changes that had to go into CMSSW are already committed directly to the vallot:cmssw.
Details can be found from the CATToolsTopAnalysis twiki page, https://twiki.cern.ch/twiki/bin/view/CMS/CATToolsTopAnalysis

Here are the major changes in this PR
- Lumi JSON file and pileup weights (June 30)
- New JEC v6 + uncertainty txt file (July 5)
- New JER scale factors (June 29)
- Pick most recent global tag, necessary to process reHLT or withHLT MC samples
- Preliminary CSV weight (July 5)

There is a fix to the cat::Jet dataformat, the default JER value will be 1.0, so users can safely multiply fJER for the real data.


jhgoh:
#577

PR #591, opened 2016-07-06
Apply JER in the ttbb analyzer

YoungKwonJo:
##


jhgoh:
expecting one more commit for bugfix. (from direct conversation)

PR #592, opened 2016-07-06
use if-statements just to avoid warning for ttbb analyzer

jhgoh:
logic should be same, just to avoid warning.


jhgoh:
adding a bugfix I made :-0

PR #595, opened 2016-07-22
draw tool update

tt8888tt:
drawing tool update(offsets and labeling... etc.)
and minor bug on dilepton analyzer fixed


jshlee:
@tt8888tt - is this the latest version of the code?

tt8888tt:
not really. I will push latest version soon

PR #614, opened 2016-09-28
Modified the trigger results in trigger-filters for dilepton channels

quark2:
Modified the primary trigger result in trigger-filters for MUEL and ELEL and ELEL channel, as HLT2 and add a secondary trigger result (for HLT) in the filters.


quark2 (commented):
There is a mistake of an indent, and it is fixed

jshlee:
+1

PR #617, opened 2016-10-20
Syc80x 1

suyong-choi:
- added jet quark/gluon likelihood variable in cat::Jet (relevant for AK4PFCHS jets only).
- Added cat::FatJet class (AK8PFCHS jets for now)
- updated python configurations to reflect new information
- added txt files for AK8 jets in data/JER
- code clean up 
- analysis plugin for FatJet variables


jshlee:
since there isnt a 1 to 1 conversion of pf jets to puppi jets, i think its better to have a separate puppi fat jet collection.

suyong-choi:
Currently there are no Puppi jets with AK8 in the miniAOD. It's probably run, but not stored, so that's probably why the partial puppi jet information is stored as userfloat.

details 
https://twiki.cern.ch/twiki/bin/view/CMSPublic/WorkBookMiniAOD2016#Jets

jshlee:
+1

jshlee:
will merge

PR #618, opened 2016-10-27
Validation workflow 76X

jhgoh:
Completing the (histogram based) validation workflow project.

By completing full workflow, one can obtain data-MC comparison plots and they will be shown in a dedicated web page.
An example page: http://higgs.hanyang.ac.kr/CATTools/#v7-6-6_TTLL

Forward port of this PR will be prepared.


jshlee:
+1

PR #621, opened 2016-10-31
Validation workflow 80X

jhgoh:
Port of #618 with Lepton+Jets channel extension.
Basic control plots are published to http://higgs.hanyang.ac.kr/CATTools by running steps in the Validation/prod


jshlee:
+1

PR #625, opened 2016-11-02
Missing dictionary FatJet

jhgoh:
FatJet was missing in the classDef, was necessary to run PAT2CAT.
@jshlee , was it successful without this fix?


jshlee:
+1

PR #627, opened 2016-11-18
Codes for measurement of top quark mass by using D mesons and J/psi are updated (+addition)

quark2:
Codes for measurement of top quark mass by using D mesons and J/psi are updated.

Also, there is a gap in CATAnalysis/python/histoHelper.py; when cmsLumi.CMSLumi() is called, the second parameter is not 0 but 2 so that the lumi I give cannot be printed.

(There is a reason that the number of commits is 1. Most of updates are done in v801, and I just pasted it rawly in v802 codes.)


geonmo:
This PR modify UOS analysis code only.

PR #630, opened 2016-11-23
h2draw tools update

pseudotop:
* deleted some files
* higgs to dimuon draw tool update


jshlee:
will merge, analysis only

PR #631, opened 2016-11-23
CATTrigger dataformat

jhgoh:
This PR introduces dataformat changes

**Motivation**
The CATTriggerProducer is originally designed to save trigger bits in boolean format, but this was not enabled by default. The original TriggerResults collection have been stored and used by analyzers.
Using TriggerResults is standard way to apply event filter, but it needs few lines of duplicated codes to initialize handle, loop over trigger names, matching, etc.
Also, there are trigger paths which are not interested in most analyses, highly prescaled or even disabled.
We can give more convenient interface to access trigger results (with prescale information as was in the original design) with compact data format.

### DIFF
--- a/CatProducer/plugins/CATTriggerProducer.cc
+++ b/CatProducer/plugins/CATTriggerProducer.cc
@@ -17,70 +17,75 @@
 #include "HLTrigger/HLTcore/interface/HLTConfigProvider.h"
 #include "DataFormats/PatCandidates/interface/PackedTriggerPrescales.h"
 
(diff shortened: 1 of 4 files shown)


jhgoh:
not to be merged - dataformat size is larger than the TriggerResults.

jhgoh:
This PR is abandoned because
  * Event size with CATTrigger object is larger than the original TriggerResults products
      * map<string, int> is little bit smaller than vector<string> + vector<bool> + vector<int>, but almost negligible factor
      * Using RefToBase in edm::Run is not supported from the framework (documented in Workbook)

Issue #624, opened 2016-11-01
Recipes for MiniAOD 80X version3


jhgoh:
Need to follow up recipes to use MiniAOD 80X version3
https://twiki.cern.ch/twiki/bin/view/CMSPublic/WorkBookMiniAOD#Run2016_23Sept2016_ReReco_80X_ve

Changes will be summarized in https://twiki.cern.ch/twiki/bin/viewauth/CMS/CATToolsTopAnalysis

jshlee:
yep this will be done for the next prod v8-0-3

jhgoh:
spreadsheet is created https://docs.google.com/spreadsheets/d/1rWM3AlFKO8IJVaeoQkWZYWwSvicQ1QCXYSzH74QyZqE/edit#gid=1456963899

jshlee:
all the MCs need to be updated

jhgoh:
I think most of recipes (up to now) which have to be applied at the production level are done.
MCs are to be updated later.

jshlee:
we can start with RD prod

jshlee:
i would like the quark gluon ID to be updated if needed.

jhgoh:
catTrigger producer does not save trigger bits for now. BadMuon/BadChargedHadron filters are not saved.

jshlee:
why not??

jhgoh:
MET filter names are commented out in catTrigger producer cfi.
https://github.com/vallot/CATTools/blob/cat80x/CatProducer/python/producers/triggerProducer_cfi.py
Instead of catTrigger producer, we are keeping the original TriggerResults. (no information loss)

BadMuon/Ch.Hadron filters are produced with EDProducers, 'keep bool_BadMuonFitler_*_*' is missing in catEventContent_cff.py

jshlee:
original TriggerResults was quite large

jhgoh:
We are keeping original TriggerResults anyway. Switch to catTrigger?

```
edmDumpEventContent test/catTuple.root 
Type                                  Module                      Label           Process   
--------------------------------------------------------------------------------------------
edm::TriggerResults                   "TriggerResults"            ""              "HLT"     
double                                "fixedGridRhoFastjetAll"    ""              "RECO"    
edm::TriggerResults                   "TriggerResults"            ""              "RECO"    
pat::PackedTriggerPrescales           "patTrigger"                ""              "RECO"    
edm::TriggerResults                   "TriggerResults"            ""              "CAT"     
[...]

PR #632, opened 2016-11-24
QG likelihood, drop partontop at production level

jhgoh:
**Motivation**
Q-G likelihood assignment for catJets.
Previous version had placeholders, but the values were not filled.

**Changes**
  * Split configuration to separate cff file, enableQGLikelihood() function.
  * Insert missing DB file for v2b
  * ESSource related configurations

**Other changes**
  * partontop is dropped, they can be produced on the fly from prunedGenParticles.


jhgoh:
QGL values are filled in.
Note: FatJets, puppi jet QGLs should not be created until supported.

jhgoh:
#624

Issue #633, opened 2016-11-25
invalid p4 in electrons


jhgoh:
some electron have invalid four momentum, nan
Can be from smearing.

jhgoh:
nan occurs for the soft electrons, pT < 5GeV where the smearing does not work.
discussion and solution from HN: https://hypernews.cern.ch/HyperNews/CMS/get/egamma/1682/2/1/1/1.html

A PR is under preparation, fall back to original p4 when nan appears.

jshlee:
can we just drop pT < 5GeV electrons? does anyone use this?

jhgoh:
No one, as far as I know. Jpsi anlaysis?

jshlee:
nope
electrons are not used to make Jpsi

jhgoh:
Closing this issue. pt cuts will be discussed in different issue.

PR #635, opened 2016-11-25
Electron nan protection, fix job submit fix

jhgoh:
**Changes**
  * Fall back to unsmeared electron momentum if nan is appeared in smeared momentum
  * Use of crabAPI in job submit script was not working properly. submit jobs using temporary file and os.system command


jhgoh:
#633

PR #641, opened 2016-12-01
updated hep-tools

jhgoh:
Updated hep-tools subpackage.

runLumiCalc, which is the wrapper of the brilcalc is updated to run at off-sites like kisti.
- Note1: The script installs necessary python virtual environment in your home directory (installation image file size is ~200MBytes)
- Note2: From the offsites, it uses web-cached database (brilcalc -c web). brilcalc can fail due to unicode problem, a fix on the brilcalc api have to be done by user.


jhgoh:
This PR does nothing on the CATTools. merging it myself.

PR #642, opened 2016-12-01
Keep first primary vertex which fails quality cut

jhgoh:
Current CATVertexProducer saves number of PV, number of good PV and the first good PV
With this PR, the producer will also keep the first PV before checking the quality cuts, to do a synchronization exercise.

If the first PV passes quality cuts, the contents will be exactly same as before. 
If the first PV fails cuts, keep the first good PV at front and append the first bad quality to the vertex collection.

- Case1-1: pv0 passes quality cut, pv1 passes quality cut... pvN passes quality cut
  - keep the pv0 only, catVertex.size() is 1. (same with the current version)
- Case1-2: pv0 passes quality cut, pv1 fails quality cut,...
  - keep the pv0 only, catVertex.size() is 1. (also same with the current version)
- Case2-1: pv0 fails quality cut, pv1 passes quality cut
  - keep pv1 at the first and pv0 at the second. catVertex.size() becomes 2.
- Case2-2: pv0 fails, pv1 fails, ... pvN passes quality cut
  - keep pvN at the first and pv0 at the second. catVertex.size() is 2.
- Case3: all pv fails
  - keep pv0 only. catVertex.size() is 1, of course nGoodPV=0.

### DIFF
--- a/CatProducer/plugins/CATVertexProducer.cc
+++ b/CatProducer/plugins/CATVertexProducer.cc
@@ -43,20 +43,31 @@ void cat::CATVertexProducer::produce(edm::Event & iEvent, const edm::EventSetup 
 
   const int nPV = recVtxs->size();
   int nGoodPV = 0;
-  std::auto_ptr<reco::VertexCollection> out(new reco::VertexCollection());
 
-  // only save the first good vertex!
-  for ( auto &vtx : *recVtxs ){
-    if ( vtx.ndof() > minNDOF_ &&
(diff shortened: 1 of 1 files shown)


jhgoh:
This PR should be transparent for most of users who is taking the nPV and nGoodPV.
If an user blindly takes catVertex[0] and does not require nGoodPV cut, it may cause selecting 'bad' PV as the reference PV. But this case is extremely rare (every single PV should fail the selection cuts).

PR #644, opened 2016-12-15
Mini isolation

jalmond:
- Add one new isolation variable to muon+electron. 
- Change txt file path when initialising rochester code. Currently this was only reading in text file if job was run in CATAnalyzer/src/ dir and if not the code runs ok but the rochester corrections for mc where not correct.
- Add some lines to MET.h to read JET systematics


jshlee:
hi @jalmond is getMiniRelIso different for electrons and muons?
If not, i think it would be better to have a common class for them.
I assume there will be this function in official cmssw, i think it would be better to use that if that exists.

jalmond:
The one difference is that one uses effective area correction and one uses delta beta corrections. I can make a common class, but i thought it was easiest to just add a new function to each producer and add a similar variable for mini isolation like the isolation that is already in cattuples. I have not seen an official cmssw function but I agree if this exists that would be better to use. I was just given this code by a colleague https://github.com/manuelfs/CfANtupler/blob/master/minicfa/interface/miniAdHocNTupler.h

John

On 20 Dec 2016, at 07:31, Jason Lee <notifications@github.com<mailto:notifications@github.com>> wrote:


hi @jalmond<https://github.com/jalmond> is getMiniRelIso different for electrons and muons?
If not, i think it would be better to have a common class for them.
I assume there will be this function in official cmssw, i think it would be better to use that if that exists.

CatProducer/plugins/CATElectronProducer.cc
```diff
@@ -165,11 +240,20 @@ cat::CATElectronProducer::produce(edm::Event & iEvent, const edm::EventSetup & i
...
+
+
+
     double elEffArea03 = getEffArea( 0.3, scEta);
     double chIso03 = aElectron.chargedHadronIso(0.3);
     double nhIso03 = aElectron.neutralHadronIso(0.3);
     double phIso03 = aElectron.photonIso(0.3);
     aElectron.setrelIso(0.3, chIso03, nhIso03, phIso03, elEffArea03, rhoIso, ecalpt);
+    aElectron.setrelMiniIso(getMiniRelIso( pfcands, dynamic_cast<const reco::Candidate *>(&aElectron), 0.05, 0.2, 10., rhoIso,elEffArea03));
+
```
jhgoh, line 249: for simplicity - you don't need dynamic cast here

DataFormats/interface/Electron.h
```diff
@@ -33,6 +33,8 @@ namespace cat {
       if( dR < 0.35) return relIso03_;
       else return relIso04_;
     }
+
+    float MiniRelIso() const {return relMiniIso03_;}
```
jhgoh: minor fix: MiniRelIso() -> miniRelIso() to keep convention

DataFormats/interface/MET.h
```diff
@@ -31,6 +31,14 @@ namespace cat {
     float unclusteredEnSumEt(int dir) const { if (dir > 0) return unclusteredEnUp_sumet_; else return unclusteredEnDown_sumet_;}
     float unclusteredEnPt(int dir) const { return hypotf(unclusteredEnPx(dir), unclusteredEnPy(dir));}
     float unclusteredEnPhi(int dir) const { return std::atan2(unclusteredEnPy(dir), unclusteredEnPx(dir));}
+
+    float JetEnPx(int dir) const { if (dir > 0) return jetEnUp_px_; else return jetEnDown_px_;}
+    float JetEnPy(int dir) const { if (dir > 0) return jetEnUp_py_; else return jetEnDown_py_;}
+    float JetEnSumEt(int dir) const { if (dir > 0) return jetEnUp_sumet_; else return jetEnDown_sumet_;}
+    float JetResPx(int dir) const { if (dir > 0) return jetResUp_px_; else return jetResDown_px_;}
+    float JetResPy(int dir) const { if (dir > 0) return jetResUp_py_; else return jetResDown_py_;}
+    float JetResSumEt(int dir) const { if (dir > 0) return jetResUp_sumet_; else return jetResDown_sumet_;}
```
jhgoh: what if user put `JetEnPx(0)`?

DataFormats/interface/Electron.h
```diff
@@ -78,6 +83,7 @@ namespace cat {
     float smearedScale_; // smearedScale = (pt_smeared)/(pt_original). Undo smearing by pt()/smearedScale()
     float relIso03_;
     float relIso04_;
+    float relMiniIso03_;
```
jhgoh: Maybe we can move this(or these) variable(s) to cat::Lepton.

CatProducer/plugins/CATElectronProducer.cc
jalmond: I simply copied the function from some code I was given. I removed the dynamic cast. I will make a new pull request.

DataFormats/interface/Electron.h
jalmond: I have moved this to Lepton.h and changed the name. I will make new pull request.

DataFormats/interface/MET.h
jalmond: I assumed since this was to get a  systematic change in MET no one would use JetEnPx(0). I changed it so that you have to enter 1. or -1. to get the up/down systematic shifts. Anything else will return 0.

DataFormats/interface/Electron.h
jalmond: I will move this to cat::Lepton.

jalmond:
Hi @jhgoh I think i made the changes you commented on. Please let me know if there is anything i missed.

jhgoh:
Should I merge this PR?

jshlee:
@jhgoh - if your ok with it, i think this can be merged

PR #645, opened 2016-12-22
use common input root file

jhgoh:
Added a cff file to define common set of test input root files, both MiniAOD (from the xrootd service) and catTuple.

"sig"(inclusive ttbar), "bkg"(DYJets) and "data"(DoubleEG) are predefined in the cff file, therefore analyzers can take the root file with the following syntax.

`from CATTools.Validation.commonTestInput_cff import commonTestCATTuples
process.source.fileNames = commonTestCATTuples["sig"]`


jhgoh:
This does not affect central production workflow, changes are localized only for testing by users.

PR #647, opened 2016-12-22
Validation modue, FCNC anlaysis, merge hep_8x

jhgoh:
Massive update on analysis related code updates

This is analysis-only PR, updates are:

- Introduce new modules for FCNC analysis synchronization
- Use updated recipe for the TTLL/TTLJ validation
- Merge hep_8x branch to cat8x to use TopAnalyzer
- New EDProducer module to compute CSV weight

There is a code change in JetProducer to use updated JetID in high eta (eta > 2.7), does not affect standard top quark analysis


jhgoh:
This is analysis-only PR, which does not change default behaviour of the workflow.
Changes in TopAnalyzer is only for code indentation level.

PR #648, opened 2016-12-23
remove jet cleaning

dygyun:
Jet cleaning was not working cause of logical error. 
so it's removed.

### DIFF
--- a/CatAnalyzer/plugins/TopAnalyzer.cc
+++ b/CatAnalyzer/plugins/TopAnalyzer.cc
@@ -435,25 +435,6 @@ TopAnalyzer::analyze(const edm::Event& iEvent, const edm::EventSetup& iSetup)
      bool pass = std::abs(jet.eta()) < 2.4 && jet.pt() > 30 && jet.LooseId() ;
      if (!pass ) continue; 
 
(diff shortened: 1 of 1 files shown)


jhgoh:
The TopAnalyzer module have been used to produce ttbar lepton+jets anlaysis ntuple, main users are HYU and Korea U.
Recently, we discovered this analyzer had a bug, the lepton used in jet cleaning can be different from the one in the analysis. From the logic, soft electrons are preferred to be picked up. The effect have been observed from Dooyeon's control plots (jet multiplicity, etc), as she had to redo the jet cleaning in her ntuple analyzer code.

Since the effect in the final result is not easy to be predicted, we concluded that removing jet cleaning in this module is most straightforward approach to solve the problem.

This PR does not affect any of central workflow, therefore it is safe to be merged.

monttj:
I think it is better to fix so that jet cleaning is done at the ntuplizer level. Otherwise, all variables NJet, NBJet.. are meaningless.

jhgoh:
This was already fixed by commit f170e8143d19d49feaa870ad407bacb6424f6e14 by Taejeong. Closing this PR.

PR #649, opened 2017-01-02
update ttbardilep drawing tool (add error band)

tt8888tt:
I've changed 'histoHelper', Specially, 'drawTH1' has been changed which is core function of the module.
So if anyone use this, please change ur code as right form. @pseudotop, @geonmo, @quark2.
This update is for drawing additional object. (for example, error band)


jhgoh:
This PR is only for macro level, does nothing on central workflow. safe to be merged.

PR #650, opened 2017-01-09
TopAnalyzer bugfix

jhgoh:
One more bugfix in the TopAnalyzer.


jshlee:
is CommonTools/scripts suppose to be updated?

jhgoh:
scripts are additional tools and should be continuously updated, changes do not affect the production.

jshlee:
ok merge when ready

PR #651, opened 2017-01-09
Update of the ttbbLepJets code

brochero:
Good moment to update the ttbb code. Several improvements as ttbar kinematic reconstruction, optimizations and updates.


brochero:
Some of the files in this PR are not needed in the CATTools repository. To avoid useless files I will make  a new PR.

PR #652, opened 2017-01-09
CatTrigger potential bugfix

jhgoh:
The trigger name in the luminosity block and actual trigger bits seem to be inconsistent in some case, making some analyzer jobs to fail.

This PR solves the problem, trigger names are retrieved using the HLTConfigProvider at the beginLuminosityBlock function and stored in the event.
With this fix, analyzer modules can read CATTrigger objects produced on the fly.

The isProductEqual() function is implemented in the catGenWeight dataformat, to suppress warning message when reading multiple files.

### DIFF
--- a/CatAnalyzer/plugins/CATTriggerBitCombiner.cc
+++ b/CatAnalyzer/plugins/CATTriggerBitCombiner.cc
@@ -7,6 +7,7 @@
 
 #include <vector>
 #include <string>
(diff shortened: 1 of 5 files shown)


jhgoh:
The change in the dataformat does not change the member variables, therefore the change does not loose compatibility.

jshlee:
+1

Issue #613, opened 2016-09-26
b-tagging SFs with BTV HIP mitigation


YoungKwonJo:
latest b-tagging SFs with BTV HIP mitigation here.

/afs/cern.ch/user/l/lwming/public/csvSFs2016/HIP_12p9_9_7_set/ 

There were calculated using the full ICHEP datasets(BCD, 12.9/fb), and BTV HIP mitigation is applied to both Data and MC

YoungKwonJo:
/afs/cern.ch/user/l/lwming/public/csvSFs2016

jshlee:
what is the status of this???

YoungKwonJo:
https://indico.cern.ch/event/565329/contributions/2283640/attachments/1326958/1992396/BTV_HIPrepord_PPD.pdf

I also sent forward of few messages from TTH group to cat-tools@cern.ch.

jhgoh:
The "new" SF from TTH group(Nov.1) is to be retrieved from /afs/cern.ch/user/l/lwming/public/csvSFs2016/HIP_12p9_9_7_set/

jhgoh:
Recipe for the workarea setup is created. https://github.com/vallot/CATTools/wiki#2016-december-v8-0-3-in-preparation
This is not final recipe.

jhgoh:
Can be done in analysis level - to be done in parallel to the new production,.

jhgoh:
Seems to be there's updated version for 36.5fb under /afs/cern.ch/user/l/lwming/public/2017csvSFs

jhgoh:
A PR to solve this problem is made and merged. Closing this issue.

PR #655, opened 2017-01-13
update CSVWeight and helper

jhgoh:
CSV weight and its helper classes are updated.

Helper is copied from https://github.com/cms-ttH/MiniAOD/blob/master/MiniAODHelper/interface/CSVHelper.h
and the SF input root file is from /afs/cern.ch/user/l/lwming/public/2017csvSFs/


jhgoh:
#613

Issue #643, opened 2016-12-06
PostICHEP recommendations


jhgoh:
v803 is under production, but we still have remaining recipes to be applied, to be done in the next round.
https://twiki.cern.ch/twiki/bin/view/CMS/CATToolsTopAnalysis#Setup_for_2016_analysis_Moriond1

(item, source twiki page, current page version/date, done in v803 version/date)
- [x] Final 2016 JSON file. source: PdmV2016Analysis(2016-12-06,r111<-2016-11-22,r105) 
- [x] MET filter update. source:  Event filters(2016-12-01,r103<-2016-11-16,r102)
- [x] Electron smearing recipe. source: EGMSmearer(2016-12-06,r28<-2016-10-24,r23) 
- [x] JEC V10. source: Recommended JEC db table(2016-12-06,r121<-July8,r119)
- [x] 80X Photon ID. N/A Dec.7
- [x] New MC samples. N/A Dec. 7
  - https://cmsweb.cern.ch/das/request?view=list&limit=50&instance=prod%2Fglobal&input=dataset%3D%2F*%2F*Moriond17*%2FMINIAODSIM
- [x] New Deep learning based flavour tagging(?) source: https://indico.cern.ch/event/592646/contributions/2392493/attachments/1383047/2103526/BTVDeepDec.pdf

jhgoh:
```
cmsrel CMSSW_8_0_24_patch1
cd CMSSW_8_0_24_patch1/src
cmsenv

git-cms-init
git checkout -b cat80x
git-cms-merge-topic -u jhgoh:PseudoTop-TOPLHCWG2016-from8024p1
git-cms-merge-topic -u cms-met:fromCMSSW_8_0_20_postICHEPfilter
git-cms-merge-topic -u shervin86:Moriond2017_JEC_energyScales

cd EgammaAnalysis/ElectronTools/data
git clone git@github.com:ECALELFS/ScalesSmearings.git
cd -

git-cms-addpkg RecoEgamma/ElectronIdentification
git clone -n https://github.com/ikrav/RecoEgamma-ElectronIdentification.git tempEGMData
cd tempEGMData
git checkout egm_id_80X_v1
mv Spring16* $CMSSW_BASE/src/RecoEgamma/ElectronIdentification/data/
cd ..
rm -rf tempEGMData

git clone https://github.com/vallot/CATTools

scram b -j30
```

jhgoh:
[...]

PR #657, opened 2017-01-18
Use egm regression

jhgoh:
Electron/Photon regression was missing.


jhgoh:
#643

jhgoh:
merging this PR.

PR #660, opened 2017-01-31
New JEC and residual scale for electrons and photons

jhgoh:
New JEC is arrived to use new MC + rereco data, Summer16_23Sep2016AllV3
- https://twiki.cern.ch/twiki/bin/view/CMS/JECDataMC?rev=r124

In addition, the full instruction to apply residual scaling for electrons and photons are available
- https://twiki.cern.ch/twiki/bin/view/CMS/EGMSmearer?rev=41

These items need new production since it changes momentum of jets, electrons and photons.
Scaling for the EGM objects cannot be done from the catTuple level since it requires ebRecHits which is not to be kept in the catTuple.


jhgoh:
#643

jhgoh:
Merging this PR, runtests OK.

PR #663, opened 2017-02-06
Update recipes for Moriond17

jhgoh:
There are recipes to be applied before starting v805 production.
In addition to this, there are non-relevant items that can be included.

- for prod: Electron ID without isolation (descriptions below)
- for prod, forward jet anlaysis: new JetID from https://twiki.cern.ch/twiki/bin/view/CMS/JetID13TeVRun2016
- just for completeness: New JEC V4 - no change compared to V3 (first to be applied from v805 anyway)
- only for analysis level: New csv files for the b tag SF

Isolation cuts are included in the cut based electron identification.
The TopEGM suggested to include VID without isolation requirement, to enable the data driven QCD estimation.

With this PR, new cut based IDs with -noiso suffix are included during the catTuple production.

see  "Inverting isolation in VID ? How ?" section of https://twiki.cern.ch/twiki/bin/view/CMS/TopEGM#Cut_based_Electron_Identificatio**


jhgoh:
unit test OK. ready to be merged.

PR #664, opened 2017-02-07
Remove unused files which prevent crab submit

jhgoh:
There are unused files with large size.
Since the default crab3 job archive have to be less than hundreds of MBs, the file size have to be reduced.
Usually only up-to-date JEC db files and txt files are used and are not used in the analysis level, we can remove them.
The lumimask json files are same.

New tab for the v805 is added in the catGetDatasetInfo script.


jhgoh:
will be nice if others can confirm that this is safe.

jshlee:
+1 this looks fine

PR #672, opened 2017-02-23
Metjec

jalmond:


jshlee:
is this for RD only or both?
can u add in labels?

jalmond:
This is for RD and MC. I will try and add labels now.

jshlee:
does this mean MC will need to be redone?
how important is this?

jalmond:
I think the effect will be small. It is just updating the JEC used in the MET. For that it is not urgent.  I also saved the xy shifted met for our trilepton analysis. It would be nice to have this in the MC also but It can wait i guess. It depends on whether we expect to have a new MC production anytime soon.

jhgoh:
I think we need new campaign, say v806 to introduce the dataformat change.

We already have some catTuples with v805 setup. Using samples with and without new variables can introduce crash when used.

jshlee:
yep im good with v806

jhgoh:
Merging this PR

PR #675, opened 2017-03-06
Simplified validation workflow

jhgoh:
Simplified the validation workflow, no need to produce syst. unc variations.


jhgoh:
Some additional commits to come later.

PR #677, opened 2017-03-14
Simplify validation #2

jhgoh:
This PR is continued from the #675.
Scripts for the 1st step "common" validation is updated, applying changes from the directory renaming, etc.


jhgoh:
Merging this PR, this only changes the validation workflow and no effect on production nor general user analyses.

PR #680, opened 2017-03-16
Disable catGenTops for TTTT sample

jhgoh:
CATGenTopProducer is not to be included in the FourTops sample.


jhgoh:
This PR does not change standard workflow. Safe to be merged.

PR #683, opened 2017-04-03
lots of updates to analysis

jshlee:
ttll and h2mu updated for 2016


jhgoh:
Are the txt files for the Rochester correction? Can it be given in different format with smaller file size?

jshlee:
i assume so... but i would rather use it as is...

jshlee:
the Rochester Correction data files were 20mb which is too large so i removed them

jshlee:
merged since this is analysis only

PR #687, opened 2017-04-24
Updated h2muAnalyzer.cc 

SangMLee:
Modified the cut for ggF Tight category so that it accepts all valid combinations that passes the preselection process like the VBF Tight category. This corrected the overall distribution of the 5 categories to closer match the synchronization exercise.


jshlee:
analysis only
will merge

PR #691, opened 2017-05-01
Histogram analyzer for FCNC

jhgoh:
Updates on FCNC histogram analyzers.
Output plots with reasonable binning, legends and labels
Additional plots for the QCD estimation


jhgoh:
Analysis only. merging.

Issue #699, opened 2017-06-11
v807 production recipe


jhgoh:
The production recipe is to be simplified

jhgoh:
The recipe is being updated. https://github.com/vallot/CATTools/wiki/Campaigns-80X#2017-june-v8-0-7

jhgoh:
DeepFlavour was not actually stored in 806.
https://twiki.cern.ch/twiki/bin/view/CMS/DeepFlavour

For the release later than 8027, tagger name is changed
deepFlavourJetTags -> pfDeepCSVJetTags
deepFlavourCMVAJetTags -> pfDeepCMVAJetTags. 

Additional recipe for the CMSSW
```
git-cms-addpkg RecoBTag/DeepFlavour
mkdir -p RecoBTag/DeepFlavour/data
wget http://home.fnal.gov/~verzetti//DeepFlavour/training/DeepFlavourNoSL.json -O  RecoBTag/DeepFlavour/data/DeepFlavourNoSL.json
wget http://mon.iihe.ac.be/~smoortga/DeepFlavour/CMSSW_implementation_DeepCMVA/Model_DeepCMVA.json -O RecoBTag/DeepFlavour/data/Model_DeepCMVA.json
```

Done in #717

jhgoh:
Dataset information:
  * Data reprocessing information: https://twiki.cern.ch/twiki/bin/view/CMS/PdmVDataReprocessing
    * Latest rereco: https://twiki.cern.ch/twiki/bin/view/CMS/PdmVDataReprocessing8029Run2016Legacy
    * Release=CMSSW_8_0_29, GT=80X_dataRun2_2016LegacyRepro_v4
  * Lumi information: https://twiki.cern.ch/twiki/bin/view/CMS/TWikiLUM
    * brilcalc lumi --normtag /afs/cern.ch/user/l/lumipro/public/Normtags/normtag_PHYSICS.json -i [yourjson]
    * Luminosity: 35.922fb-1, uncertainty 2.5%

MC samples (unchanged from the previous prod)
  * https://twiki.cern.ch/twiki/bin/view/CMSPublic/WorkBookMiniAOD#Run2_Moriond17_re_digi_reco_camp
  * Release>=8021, GT=80X_mcRun2_asymptotic_2016_TrancheIV_v6

jhgoh:
Pileup information (unchanged from the last prod)
[...]

PR #702, opened 2017-06-16
remove dataset submodule

jhgoh:
dataset txt file is not well maintained, usually list is retrieved with catGetDatasetInfo.
Removing this will simplify recipe.


jhgoh:
#699

PR #705, opened 2017-06-19
Refactoring with CommonTools 

jhgoh:
This PR is for the coding style - reduce code duplicates, migration to unique_ptr, remove unused variables and functions, etc.

Modules which do not have direct dependency to the CAT dataformat are moved to CommonTools.

There should no physics change with with PR.

I expect some level of enhancement in CPU usage in CATElectronProducer and CATMuonProducer - getByToken is taken out from the lepton loop, genParticles are selected before the MC matching loop.


geonmo:
+1

jshlee:
+1

Issue #606, opened 2016-08-10
File size reduction


jhgoh:
catTuple size is increasing compared to the early period.
The total size is manageable for now, but it will be a serious problem at the end of the Run-II.

jhgoh:
For the 800 production (selecting items >~100Bytes/Event):
- DYJets MC background

```
File /xrootd/store/group/CAT/DYJetsToLL_M-50_TuneCUETP8M1_13TeV-amcatnloFXFX-pythia8/v8-0-0_RunIISpring16MiniAODv2-PUSpring16_80X_mcRun2_asymptotic_2016_miniAODv2_v0-v1/160705_214259/0000/catTuple_1.root Events 38517
Branch Name | Average Uncompressed Size (Bytes/Event) | Average Compressed Size (Bytes/Event) 
recoGenParticles_prunedGenParticles__PAT. 4442.01 735.487
catJets_catJets__CAT. 2512.01 350.633
catSecVertexs_catSecVertexs__CAT. 1207.5 331.873
catGenWeights_genWeight__CAT. 488.546 254.201
recoGenJets_slimmedGenJets__PAT. 849.908 165.87
catMuons_catMuons__CAT. 406.488 139.88
catJets_catJetsPuppi__CAT. 892.033 132.696
catSecVertexs_catDstars_JpsiMVACand_CAT. 412.518 105.71
catMETs_catMETsPuppi__CAT. 265.228 103.381
catMETs_catMETs__CAT. 265.149 102.304
patTriggerObjectStandAlones_catTrigger__CAT. 2094.79 99.2689
catSecVertexs_catDstars_D0Cand_CAT. 444.836 92.6301
catElectrons_catElectrons__CAT. 647.507 85.4004
```
- DoubleEG real data

```
File /xrootd/store/group/CAT/DoubleEG/v8-0-0_Run2016B-PromptReco-v2/160705_213824/0000/catTuple_1.root Events 113638
Branch Name | Average Uncompressed Size (Bytes/Event) | Average Compressed Size (Bytes/Event) 
catSecVertexs_catSecVertexs__CAT. 1422.52 390.143
catJets_catJets__CAT. 2726.52 309.702
catSecVertexs_catDstars_D0Cand_CAT. 719.567 157.212
catSecVertexs_catDstars_JpsiMVACand_CAT. 505.863 130.709
catElectrons_catElectrons__CAT. 898.711 118.442
[...]

PR #706, opened 2017-06-19
Reduce event size for bkg MC

jhgoh:
This PR reduces event size of catTuple for the bkg MCs.

A flag "isSignal" is added to drop genParticles in the output event content, also drop genWeights for the systematic uncertainty variations.

Deciding the "signal" and "background" definition is up to the users.

With this PR, we can save
- 785.628 Bytes/Event by dropping prunedGenParticles
- 326.873-28.752 = 298.121 Bytes/Event by keeping the nominal genWeight only

which is 20% of total size for the DY sample.

### DIFF
--- a/CatProducer/plugins/GenWeightsProducer.cc
+++ b/CatProducer/plugins/GenWeightsProducer.cc
@@ -182,6 +182,7 @@ private:
   const bool enforceUnitGenWeight_;
   const bool doLOPDFReweight_;
   bool reweightToNewPDF_;
+  const bool doKeepFirstOnly_;
   const edm::EDGetTokenT<LHEEventProduct> lheToken_;
   const edm::EDGetTokenT<GenEventInfoProduct> genInfoToken_;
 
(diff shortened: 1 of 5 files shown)


jhgoh:
#606

jhgoh:
Can we drop genJets for the bkg as well?

jshlee:
i assume so

jhgoh:
PR is updated
- Drop GenJets for bkg MC
- Do not keep flags, as they are available in the catTrigger

jshlee:
i think keeping them default would be better for testing by others
for prod, its easier to turn off

jshlee:
how should we determine signal samples???

jhgoh:
"isSignal" can be set in the spreadsheet, use "options" tab (separate with commas)

geonmo:
@jshlee Could you final review for this PR before next CAT production?

geonmo:
@jhgoh However, I confused to determine bkg sample. For example, DY sample is a bkg for top quark user but is a signal for DY users.  To choose signal sample  should be judged carefully.

jhgoh:
@geonmo yes that was one of the message, but are there anybody use DY as signal and do the PDF systematics?

jshlee:
+1

jshlee (approved):

jhgoh:
I'm merging this. The default options are set to be conservative, default setup will keep all weights.

PR #709, opened 2017-07-04
Updated of ttbb->Lep+Jets

brochero:

PR #710, opened 2017-07-04
Remove code duplication in KinFit

jhgoh:
This PR is a technical change.
  * Removes code duplications in KinFitter routines.
  * Global variables are declared as "extern", so that the header file can be included multiple times (TopAnalyzer and ttbbLepJetsAnalyzer can use the same routine).


jhgoh:
Test done, identical result except the change in the b tag WP.
image

Red(current) and blue(new with b tag WP CSV>0.800) is identical,
but the WP 0.800 is obsolete, therefore it has to be updated to CSV>0.8404 or different WP likewise done in the LepJetsFitter.h by @brochero #709

jhgoh:
Merging this PR. 
Note: FCNH fitter is to be optimized based on future study.

PR #711, opened 2017-07-05
PartonTop tau decay + FCNC

jhgoh:
PartonTopProducer is updated,
- Better treatment of tau decays, fixed tau->gamma+X/gamma->e+e- (less than 0.1%)
- Some more extension to top FCNC via Z
- Technical changes including coding style.


jhgoh:
Test done, runs without problems. merging this PR.

PR #716, opened 2017-08-07
Keep jet constituent pt/eta/phi

jhgoh:
This PR keeps jet constituents' pt,eta,phi in vector<array> to cover the jet imaging study.

Event size is expected to be increased with this change, but it should be better than saving all pf candidates.

Below are the size of cat::Jet before and after this update (x2 increase of cat::Jet)
```
        Uncompressed Compressed
Before       8585.63   1225.23
After        9978.81   2297.51
```

### DIFF
--- a/CatProducer/plugins/CATJetProducer.cc
+++ b/CatProducer/plugins/CATJetProducer.cc
@@ -205,6 +205,11 @@ void cat::CATJetProducer::produce(edm::Event & iEvent, const edm::EventSetup & i
     int partonPdgId = aPatJet.genParton() ? aPatJet.genParton()->pdgId() : 0;
     aJet.setPartonPdgId(partonPdgId);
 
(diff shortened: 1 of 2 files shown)


geonmo:
@jhgoh What is test sample? TTJet? or QCD?

jhgoh:
@geonmo it's TTJets

jshlee:
i think this study is best to be done on miniAOD or AOD

geonmo:
@jshlee Could you recommend members to test for miniAOD and AOD after KPS? I think that it is easy and interesting to study CATTools development.

jshlee:
I still think its better to do this study at miniAOD or AOD level.

jhgoh:
Closing this PR, I guess we never have plan to have this feature in 10x and future.

PR #720, opened 2017-11-26
Skim CATTuple production

jhgoh:
Skim event, can apply lepton/jet multiplicity cuts.
Loose selections on leptons and jets, considering energy resolution therefore will be safe to do the systematics.

crab configuration is modified little bit as well;
- Changed output path, save under /store/group/CAT/vX-Y-Z/Dataset/etc/etc
- Do not apply JSON to process real data, FileBased splitting

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ __init__.py
 .#*
 #*#
 *~
+crab_v*
+crab.log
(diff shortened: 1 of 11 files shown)


jhgoh:
catGetDatasetInfo to be updated after testing submission/production procedure.

jhgoh:
Merging this PR, as mentioned during the meeting.

PR #730, opened 2018-05-28
cattuple production for 31Mar2018 ReReco

minerva1993:
This is a PR for CatTuple production for 31Mar2018 ReReco 2017 dataset and corresponding MC samples.

-Update B tag SF
-Create and implement deepCSV & its weight
-update ID, Iso, track SF for lepton
-Jet ID update
...and so on.

An analyzer and config file for FCNC search is also added.

Thanks,
Jiwon Park


CatProducer/scripts/catGetDatasetInfo
```diff
@@ -138,7 +138,7 @@ for d in ds:
     if len(dataSetName) < 4: continue
 
     pdName, sdName = dataSetName[1], dataSetName[2]
-    if [int(x) for x in era[1:].split('-')] >= [8,0,8]:
+    if [int(x) for x in era[1:].split('_')] >= [9,1]:
```
jhgoh: have to be kept as is.

DataFormats/interface/Electron.h
```diff
@@ -87,6 +92,7 @@ namespace cat {
     
     int snuID_;
     bool isTrigMVAValid_;
+    bool isHLTEle32WPTight_;
```
jhgoh: Adding a member variable just for one era is not a good idea in general, you will loose backward / forward compatibility.

DataFormats/src/GenTop.cc
```diff
@@ -117,7 +117,27 @@ void GenTop::building(Handle<reco::GenJetCollection> genJets, Handle<reco::GenPa
...
+        if ( decayId == 5 ) {
+          Hbquarks_[nHbquarkDaughters] = decay->p4();
+          nHbquarkDaughters++;
+        }
+      }
+      continue;
+    }
+    // end of Singletop FCNH side
+    */
+
```
jhgoh, line 140: necessary to this part?

CatAnalyzer/prod/job_fcnc.sh
```diff
@@ -0,0 +1,87 @@
+./create-batch --jobName SingleMuon_Run2017B  --fileList $CMSSW_BASE/src/CATTools/CatAnalyzer/data/dataset/dataset_SingleMuon_Run2017B.txt  --cfg fcncAnalyzer_cfg.py --transferDest /store/user/minerva1993/ntuple_jw/2017/V9_2/production/SingleMuon_Run2017B --maxFiles 30 --args 'UserJSON=true' 'runOnTTbarMC=0' 'TTbarCatMC=0'
```
jhgoh: do we have create-batch in the prod directory?

CatProducer/scripts/catGetDatasetInfo
minerva1993: T.J. wanted to change production tag, like 'V9_1'. https://docs.google.com/spreadsheets/d/1rWM3AlFKO8IJVaeoQkWZYWwSvicQ1QCXYSzH74QyZqE/edit#gid=1986226361

CatAnalyzer/prod/job_fcnc.sh
minerva1993: Hmm, this is temporary solution to deal with some missing lines from V9_1 recipe (about submodule). We can add those lines in wiki later

DataFormats/src/GenTop.cc
minerva1993: This is a subtle point, since v80x GenTop cannot store Higgs from single top FCNC samples. Quick solution that I used is copy those lines for single top FCNC production only, rather than trying to recompose GenTop codes.

CatProducer/scripts/catGetDatasetInfo
jhgoh: With the proposed change, script will crash if someone tries to retrieve 80X release info.
rather, I would suggest
```
if [int(x) for x in era[1:].replace('_', '-').split('-')] >= [8,0,8]:
```
This seems to be working.

minerva1993: Yes, it seems working fine.

DataFormats/interface/Electron.h
minerva1993: I and T.J. discussed about this issue.
First of all, this seems easiest way to check if trigger path (HLT_Ele32_WPTight_L1Double) passed singleEGOr L1 filter.
Secondly, people are recommended to use proper cattools and cmssw version to open and analysis catTuples. Even if the content of this var & function change in future, it would be okay.

Also, T.J. suggested that we may change the name of var and function to general one (let's say, 'isTrgPass' or something like that), so we can put some useful information on future production (in case of 2018 or later data).

jhgoh:
Ready to be merged?

minerva1993:
@jhgoh  Yes, I think so

PR #733, opened 2018-09-09
Add EE noise mitigation for MET

BrieucF:
- Based on https://twiki.cern.ch/twiki/bin/viewauth/CMS/MissingETUncertaintyPrescription#Instructions_for_9_4_X_X_9_for_2 
- Minimum release version required is CMSSW_9_4_9
 - Need to add `git cms-merge-topic cms-met:METFixEE2017_949` to the recipe
 - Must be used with the latest JEC (`Fall17_17Nov2017_V8`) which is already included in GT so it should be ok
 
Impact on MET (red is with mitigation, blue without) for MC:

slimmemed_rediswithcorrection

Impact on MET (red is with mitigation, blue without) for data:

slimmedmet_rediswithcorrection_data

NB:
 - an update of this recipe should be delivered soon by JETMET
 - this is a temporary fix, a proper modeling of the EE noise will be included in next re-reco

### DIFF
--- a/CatProducer/python/catTools_cff.py
+++ b/CatProducer/python/catTools_cff.py
@@ -16,7 +16,7 @@ def catTool(process, runOnMC=True, useMiniAOD=True):
             LumiSections = LumiList('%s/src/CATTools/CatProducer/data/LumiMask/%s.txt'%(os.environ['CMSSW_BASE'], cat.lumiJSON)).getVLuminosityBlockRange())
 
         #process.load("CATTools.CatProducer.eventCleaning.badECALSlewRateMitigationFilter2016_cfi")
-    
+
 #    useJECfile = True
 #    jecFiles = cat.JetEnergyCorrection
 #    if runOnMC:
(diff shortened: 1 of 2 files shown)


jhgoh:
Basically, https://github.com/vallot/CATTools/blob/cat90x/setup.sh is the file to be modified to address git-cms-addpkg

minerva1993:
From the setup script, the last line would be catGetDatasetInfo v9_3 (for new production).
And when we checkout the branch, we will use a specific tag (again, V9_3) for the new production.

And from https://github.com/vallot/CATTools/wiki/Recipe-for-V9, we may need to add
git cms-addpkg PhysicsTools/PatAlgos
as well (I'm not sure).

BrieucF:
I think you are right, we have to add PatAlgos (in any case it does not hurt). I will put this in the setup.sh for consistency. For the V9_3, it is not created yet, we could update setup.sh once we think everything is ready for next prod and once we have created the v9_3 tag.

PR #734, opened 2018-09-17
Bugfix scale factor evaluator

jhgoh:
Bugfix for the ScaleFactorEvalulator.
Fixes bin shifting in some cases. 

We were using std::lower_bound to find histogram binning, but actually the name was given to return the "lower boundary of elements with the same value" - to return 4th element if "44" was asked to be found in a vector [0,1,2,44,44,44,555,678]. And the actual implementation is done to do the binary search with the comparison operator.

Simple and correct solution is to return the element BEFORE std::upper_bound().


minerva1993:
I just tested with 2017 TTLJ samples with corresponding SFs, and it seems to work properly. By the way, did you mean #include <algorithm> in header? (not algorithms)

jhgoh:
I'll merge this PR tonight if there's no objection

PR #740, opened 2018-12-04
Development for v9 4

minerva1993:
Several updates for cattools V9_4.

Following histos are plotted with TTLJ MiniAOD v2 sample, 1000 events.

1) b tagging SF: updated for working point based SF. Shape correction is unchanged.  Red line is V2, dashed line is V3.
bsfv3

2) Egamma VID v2: "when the 94X DY+Jets samples with a more realistic pileup scenario were produced, it was decided to retrain and call it V2., Besides being trained with a more adequate MC sample, the V2 ID benefits from being trained with the fast xgboost algorithm instead of the standard TMVA gradient boosting." - https://twiki.cern.ch/twiki/bin/view/CMS/MultivariateElectronIdentificationRun2#Differences_between_V1_and_V2
Red line is V1, dashed line is V2.
egmid

3) JER updated, Fall17_V3. Dashed line is Fall17_V3 while continuous line is the latest 16 JER
jer

4) EE noise fix for MET is included in CMSSW_9_4_12. Red line is implementation in CMSSW_9_4_9 with private repository, and dashed line is 9_4_12.
met

5) Muon selector. It is strongly recommended to use muon selector instead of boolean selector. (https://twiki.cern.ch/twiki/bin/viewauth/CMS/SWGuideMuonIdRun2#Muon_selectors_Since_9_4_X). I added member functions and variables in CAT dataformat, and stored selector bit which is already computed in miniAOD. Please double check this part if works are done properly.
i) Selector bits in miniAOD
selector_miniaod

ii) Selector bits in CatTuple
selector_cattuple

iii) pT of Tight muon with PFIsoTight, pT > 20 and |\eta| < 2.4. Red line is standard boolean ID (isTight()) and dashed line is selector.
selector

6) JEC: waiting for new globaltag

7) Cleaned up outdated JER, b-SFs.


DataFormats/interface/Muon.h
```diff
@@ -58,6 +59,44 @@ namespace cat {
...
+      SoftMvaId              = 1UL<<14,  
+      MvaLoose               = 1UL<<15,  
+      MvaMedium              = 1UL<<16,  
+      MvaTight               = 1UL<<17,
+      MiniIsoLoose           = 1UL<<18,  // reliso<0.40
+      MiniIsoMedium          = 1UL<<19,  // reliso<0.20
+      MiniIsoTight           = 1UL<<20,  // reliso<0.10
+      MiniIsoVeryTight       = 1UL<<21   // reliso<0.05
+
+    };
```
jhgoh: Isn't it possible just to use those enum types already defined in the Muon.h? Usually copy & paste become painful after few months.

DataFormats/interface/Muon.h
```diff
@@ -73,6 +112,8 @@ namespace cat {
     int numberOfMatchedStations_;
     int numberOfValidPixelHits_;
     int trackerLayersWithMeasurement_;
+
+    unsigned int selectors_;
```
jhgoh, line 87: This introduces dataformat change. Increasing cattuple version name could be better.

jhgoh:
+1

DataFormats/interface/Muon.h
minerva1993: Did you mean cat91x or something like this? Yes, it will be more clear to increase version, but notice that only one group is using cattools for 2017 analysis: after finalizing tools and deleting old cattuples, tag for V9_4 will be sufficient, I guess.

DataFormats/interface/Muon.h
minerva1993: done!

minerva1993:
I would like to include JEC update as well. As soon as finalized, I will update the PR.

minerva1993:
Here's 2017 JEC update to V32. Red is older (included in global tag v14 for MC) and dashed line is JEC V32
jecpt without catskimevent

jecup

I updated catTools_cff, referring to WorkBook: https://twiki.cern.ch/twiki/bin/view/CMSPublic/WorkBookJetEnergyCorrections#CorrPatJets

Also I disabled jet part of catSkimEvent to avoid an error. If you have any idea, please let me know.

jhgoh:
catSkimEvent was never used practically. will be pruned out after this PR.

DataFormats/interface/Muon.h
jhgoh: On the other hand, why don't you use std::bitset?

BrieucF:
Sorry to arrive late on this topic but, just to be sure, are we propagating the new JES and JER to MET? Since MiniAOD was produced with old JEC's, I assume it should be re-computed?

minerva1993:
@BrieucF
Actually I forgot to turn on that part from catTools_cff, but fortunately, 'fixEE'part will do this job, since it is an option for MET correction:
https://github.com/vallot/CATTools/blob/cat90x/CatProducer/python/catTools_cff.py#L119-L128

see here:
https://twiki.cern.ch/twiki/bin/viewauth/CMS/MissingETUncertaintyPrescription

PR #743, opened 2019-01-11
retrieve cutbasedID, correct AEff, modify noiso ID

minerva1993:
Here's some codes to retrieve cunBasedElectronID. (Because the electron trigger SF made by ttH group used this ID) 

NoIso id code is modified, following
https://github.com/guitargeek/cmssw/blob/EgammaID_949/RecoEgamma/ElectronIdentification/python/Identification/cutBasedElectronID_Fall17_94X_V2_cff.py#L7-L9

and

https://github.com/guitargeek/cmssw/blob/EgammaID_949/RecoEgamma/ElectronIdentification/python/Identification/cutBasedElectronID_tools.py#L746


minerva1993:
If there is no additional comment by ~ 6 pm, I will merge this PR

PR #747, opened 2019-01-23
Cat10x cleanup

minerva1993:
Deleted: 
-Old root files
-Codes for B SF from ttH from analysis team (root file based). Use only B POG recipe.
-SF, private id in data format
-Old SFs
-Sec. Vertex and D star
-Some of old analyzer
-Validation folder
-Outdated MET filters (now included in MiniAOD as flag, stored in cat trigger)

We will receive reviews and comments by the next Wednesday. If there is anything can be removed further or the lines you want to keep, please let me know.


jhgoh:
nice. 
@sarakm0704 I don't think kinreco root files have to be kept under catttols-datasets. anybody who need root files can go back to 80x branch to retrieve them.

minerva1993:
If there is no comment, the PR will be merged by tomorrow (Jan 31st) noon.

PR #750, opened 2019-03-06
Cat80x

Somhammer:
ttbbLepJetsAnalyzer.cc: remove all kinematic reconstruction code
leptonSF_cff.py: update electron trigger SF
Gentop.cc: modify loop bugs


minerva1993:
gentop fix which is done on cat90x and 10x already, and the other one is ntuplizer modification. This PR can be merged

PR #752, opened 2019-04-30
Cat80x

Somhammer:
ttbbLepJetAnalyzer can take more than 4 gen bjets from gentop


CatAnalyzer/plugins/ttbbLepJetsAnalyzer.cc
```diff
@@ -732,10 +788,44 @@ void ttbbLepJetsAnalyzer::analyze(const edm::Event& iEvent, const edm::EventSetu
...
     b_DRAddJets = genttbarConeCat->begin()->dRaddJets();
 
+    // adding b jet four-momentum having minimum deltaR
+    std::vector<math::XYZTLorentzVector> genbjets = genttbarConeCat->begin()->bJets();
+    tmp2->Fill(genbjets.size());
+    double mindR=999;
+    int index1=0, index2=0;
+    for(unsigned int i=0; i < genbjets.size(); ++i){
+      if( genbjets[i].Pt() < 20 || std::abs(genbjets[i].Eta()) > 2.5 ) continue;
+      for(unsigned int j=0; j < genbjets.size(); ++j){
```
jhgoh: `for ( unsigned int j=i+1; j < .... )` is usually better

CatAnalyzer/plugins/ttbbLepJetsAnalyzer.cc
```diff
@@ -732,10 +788,44 @@ void ttbbLepJetsAnalyzer::analyze(const edm::Event& iEvent, const edm::EventSetu
...
+
+    b_mindRbjet1_pt = genbjets[index1].Pt();
+    b_mindRbjet1_eta = genbjets[index1].Eta();
+    b_mindRbjet1_phi = genbjets[index1].Phi();
+    b_mindRbjet1_e = genbjets[index1].E();
+
+    b_mindRbjet2_pt = genbjets[index2].Pt();
+    b_mindRbjet2_eta = genbjets[index2].Eta();
+    b_mindRbjet2_phi = genbjets[index2].Phi();
+    b_mindRbjet2_e = genbjets[index2].E();
```
jhgoh: could be very rare - this will be crashing if number of genbjets == 0. 

In addition, it can give wrong answer in some cases such as:
- there's only one genbjet -> still gives index1 = index2 = 0
- there's two genbjet but the 0th one fails acceptance -> gives index1 = 1, index2 = 0

CatAnalyzer/plugins/ttbbLepJetsAnalyzer.cc
```diff
@@ -732,10 +788,44 @@ void ttbbLepJetsAnalyzer::analyze(const edm::Event& iEvent, const edm::EventSetu
...
+
+    b_mindR = mindR;
+
     if(genttbarConeCat->begin()-> NaddbJets20() > 1){
       EventInfo->Fill(2.5, 1.0); // Number of ttbb Events	
       if(genttbarConeCat->begin()->semiLeptonic(-1)) EventInfo->Fill(3.5, 1.0); // Number of ttbb Events (Includes tau) 
-      if(genttbarConeCat->begin()->semiLeptonic(0)) EventInfo->Fill(4.5, 1.0); // Number of ttbb Events (Includes tau leptonic decay)
+      if(genttbarConeCat->begin()->semiLeptonic(0)){
+	  EventInfo->Fill(4.5, 1.0); // Number of ttbb Events (Includes tau leptonic decay)
+          tmp1->Fill(0.5,1.0);
```
jhgoh: Filling with weight=1 ... is it intended? ignore genWeight?

jhgoh: BTW, tmp1 is a bad idea for a variable name. give a meaningful name unless it is really a temporary variable used locally. I think this tmp1 starts from the very first part of this module.

Somhammer: tmp1 was made to check the events which are not categorized as ttbb but have addbjets. We already found the reason and solved, tmp1 is not anymore used. I will remove this tmp1.

CatAnalyzer/plugins/ttbbLepJetsAnalyzer.cc
```diff
@@ -732,10 +778,46 @@ void ttbbLepJetsAnalyzer::analyze(const edm::Event& iEvent, const edm::EventSetu
     // DR 
     b_DRAddJets = genttbarConeCat->begin()->dRaddJets();
 
+    // adding b jet four-momentum having minimum deltaR
+
+    std::vector<math::XYZTLorentzVector> genbjets = genttbarConeCat->begin()->bJets();
+    if( genbjets.size() > 0 ){
```
Somhammer: 1. The loop will run only when the number of genbjets is more than 0.
2. index1 and index2 are initialized as -1. And branch will be filled when index is not -1. Therefore analyzer does not return wrong answer(maybe).

CatAnalyzer/plugins/ttbbLepJetsAnalyzer.cc
```diff
@@ -732,10 +778,46 @@ void ttbbLepJetsAnalyzer::analyze(const edm::Event& iEvent, const edm::EventSetu
...
+	  double tmp = reco::deltaR(genbjets[i], genbjets[j]);
+	  if( tmp < mindR ){
+	    mindR = tmp;
+	    index1 = i;
+	    index2 = j;
+	  }
+        }
+      }
+
+      if( index1 > -1 && index2 > -1){
```
Somhammer: 1. The loop will run only when the number of genbjets is more than 0.
2. index1 and index2 are initialized as -1. And branch will be filled when index is not -1. Therefore analyzer does not return wrong answer(maybe).

CatAnalyzer/plugins/ttbbLepJetsAnalyzer.cc
```diff
@@ -732,10 +778,46 @@ void ttbbLepJetsAnalyzer::analyze(const edm::Event& iEvent, const edm::EventSetu
     // DR 
     b_DRAddJets = genttbarConeCat->begin()->dRaddJets();
 
+    // adding b jet four-momentum having minimum deltaR
+
+    std::vector<math::XYZTLorentzVector> genbjets = genttbarConeCat->begin()->bJets();
+    if( genbjets.size() > 0 ){
+      double mindR=999;
+      int index1=-1, index2=-1;
+      for( unsigned int i=0; i < genbjets.size(); ++i ){
```
Somhammer, line 787: 1. The loop will run only when the number of genbjets is more than 0.
2. index1 and index2 are initialized as -1. And branch will be filled when index is not -1. Therefore analyzer does not return wrong answer(maybe).

minerva1993:
I think we can merge this branch if anyone has objection on gentop modification which stores all gen b jets for further study.

PR #764, opened 2020-07-07
add new variables in addbJets

sarakm0704:
add variables for addbjets with selections


monttj:
Hi Jieun,

We can probably simply apply the eta cut for all of the additional jets. 
In this way, we can use the existing variable which might be the simplest way. 
Let's try to talk on Monday. 

Taejeong

PR #765, opened 2020-07-07
GenTop Modification for cat 10x

Somhammer:
1. GenTop:
  All additional b jet candidates are saved and you can access them by addbJets().
2. ttbbLepJet analyzer:
  scale uncertainties are saved for all ttbar samples.


minerva1993:
This PR is being revised, merging will be done after committing all the other fixes.

minerva1993:
It seems that the definition of ttjj from dR<0.4 between gen jet and quark can cause some problem. Let's think about this case.
- Light gen jet from W have dR > 0.4 with its mother parton.
- In that case, the gen jet stored in additional jet
- If the gen jet is within acceptance, it is counted as additional jet, may increase the number of ttjj events. 
- Currently, the event categorization for, let's say ttbj, requires the events to be ttjj as well as the number of additional b jet >= 1. If the event is tt+1b (GenTtbarId = 51 or 52) and accidentally ttjj criteria is satisfied, the event goes to ttbj. This seems not correct.

Do you have any idea?

PR #768, opened 2020-07-08
GenTop Modification for cat 81x

Somhammer:
All additional b jet candidates are saved and you can access them by addbJets().


monttj:
Hi Seohyun,

Good to have flexibility with this approach.  
But I think it would be better not to use a loop at the analysis level. This was the original idea of having cattools. 
Please take into account the Jieun's PR and try to somehow merge together. 
We can talk on Monday. 

Taejeong

PR #770, opened 2020-07-14
GenTop Modifications for cat 90x

sarakm0704:
1. Variables for addbjets with selections (addbJets20Had) are added.
2. All additional b jet candidates are saved and you can access them by addbJets().

### DIFF
--- a/DataFormats/interface/GenTop.h
+++ b/DataFormats/interface/GenTop.h
@@ -91,6 +91,9 @@ namespace cat {
       else return addbJets_[1];
     }
 
+    const math::XYZTLorentzVector addbJets20Had1() const { return addbJets20Had_[0]; }
+    const math::XYZTLorentzVector addbJets20Had2() const { return addbJets20Had_[1]; }
+
     const math::XYZTLorentzVector bJetsFromTop1() const { return bJetsFromTop_[0]; }
     const math::XYZTLorentzVector bJetsFromTop2() const { return bJetsFromTop_[1]; }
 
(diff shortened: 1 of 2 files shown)


minerva1993:
What happens if NaddbJets20BHad_ > 2? If you draw 'genconecatid[6]' in ttbb category from ttbbLepJets ntuple (=Seohyeon's ntuple), it plots NaddbJets20(). [link1](https://github.com/vallot/CATTools/blob/cat10x/DataFormats/interface/GenTop.h#L208-L210) [link2](https://github.com/vallot/CATTools/blob/cat10x/CatAnalyzer/plugins/ttbbLepJetsAnalyzer.cc#L756-L757)

You can see there're ~99650 events with 2 add b jets, but still ~2700 events with NaddbJets20BHad_ > 2. In this case, if you use addbJets20Had1/2, you may miss true add. b jets. It is >2.5%, not a small fraction and may cause difference during the synchronization with UZH. (I guess UZH don't use gen jet indices from genTtbarID since they are not stored in NanoAOD?) I understand this way is to avoid additional processing in ntuplizer but you may need to restrict abs(eta) < 2.4 and consequently run loop over addbJets() to match with genTtbarID and UZH method.

sarakm0704:
Well, from what I understand so far, aren’t we trying to use the first two leading add.bjets that passed the threshold (pT > 20 GeV, |eta| < 2.5)? But because there are more than 2.5, we would like to make new variables that are suitable for ttbb SL differential analysis.

So isn't it inevitable that things we lose will happen? By adding the new variables, there wouldn't be any effect on other analyses. And for the synchronization, we agreed with to go < 2.5.

minerva1993:
Ok, good for using |eta| < 2.5. Do other groups using same addbjet definition as ours? I mean, taking gen jet indices from genTtbarId? As fas as I know, there's no additional gen jet indices in nanoAOD so they should add those branches as well. Unless synchronization is by construction impossible. Furthermore, for internal consistency check, I strongly recommend to add GenTtbarId with |eta| < 2.5, by *adding* additional eta cut [here](https://github.com/vallot/CATTools/blob/cat10x/CatProducer/python/Tools/tools.py#L56-L68) and [here](https://github.com/vallot/CATTools/blob/168c6d06e6fb3efb312c40a065396a833245f109/CatProducer/python/catCandidates_cff.py#L53-L54). You *must not* modify default GenTtbarCategories branch but add new one, since FCNC analysis is using |eta|<2.4 selection in accordance with 2016 paper.

minerva1993:
And why don't you include the commits that Seohyeon did before (https://github.com/vallot/CATTools/pull/765/files)? If you don't exclude additional b and c jet indices when counting additional jets for ttjj, you may double count additional jets.

sarakm0704:
I took out the part about add.cjets because it would be better to push it with another commit. I can put it again or seohyun could.

PR #784, opened 2020-09-18
Modif pdfweights and Muon Iso

Somhammer:
1. Fill pdfweights in the same way in cat81x.
2. Remove muon isolation cut if doLooselepton is true.


Somhammer:
3. Modify typo in submit script.

